PR #13, opened 2019-02-26
Enforce global limit in Nitro

### LINKED ISSUE #8: Nitro: Respect global limit
The global environment limit is not currently respected in Nitro. This needs to be implemented.

bkeroackdsc:
This enforces the global environment limit using the Nitro engine. Unlike Amino, we lock around each environment destroyed in order to preempt any operations that might be running on them.

This also adds a new template variable for notification templates: `{{ .Event }}`, which is a string representing the event that triggered the notification ("CreateEnvironment" etc). The default destroy notification is changed to specify "Environment Limit Exceeded" when the destroy event is caused by enforcing the global limit.

Fixes #8


pkg/nitro/env/env.go
```diff
@@ -247,6 +248,45 @@ func (m *Manager) Create(ctx context.Context, rd models.RepoRevisionData) (strin
...
+	}
+	limit := int(m.GlobalLimit)
+	qae, err := m.DL.GetRunningQAEnvironments()
+	if err != nil {
+		return fmt.Errorf("error getting running environments: %v", err)
+	}
+	extant := len(qae)
+	if extant > limit {
+		kill := extant - limit
+		sort.Slice(qae, func(i int, j int) bool { return qae[i].Created.Before(qae[j].Created) })
```
mikeykhalil: I wonder if it would be more efficient to handle this in the datalayer? Maybe we could have a new method, say 
`GetRunningQAEnvironmentsOrderedByAge` or something along those lines, which would basically just be an SQL query with some `Order By` clause. 

Unsure if it would make much of a difference, but figure it might be worth checking out.

bkeroackdsc: Good catch. I modified the datalayer method to return a sorted list. This is the only place where `GetRunningQAEnvironments()` is used, so it didn't seem worth creating a new method or changing the signature of this one.

mikeykhalil:
Good idea adding the new notification message. 

LGTM!

PR #14, opened 2019-03-01
datadog apm: initial support

mikeykhalil:
**Summary**
The purpose of this PR is to switch from using New Relic APM to Datadog APM.

Datadog APM has`gorilla/mux` integration. This means that we can essentially get APM for all HTTP endpoints for free. However, this integration does not provide much insight when these endpoints send responses while they are still performing asynchronous actions.

So for the endpoints that do some type of actions after an initial response is sent back to the client, it is still helpful to create some spans to properly monitor the operations.

In order to support this, I slightly modified how we handle http error responses, so that we can optionally pass spans to these functions.

I also wanted to avoid making changes to `pkg/spawner`, as I believe the plan is to remove much of this once we deprecate amino. So we will need to officially remove new relic as a dependency once we deprecate amino. Until then, we simply would pass a null newrelic struct to the `NewQASpawner`.

Did a few e2e tests, but would like to verify a few more things on Monday.


cmd/server.go
```diff
@@ -338,9 +342,11 @@ func server(cmd *cobra.Command, args []string) {
 	httpapi.WaitForHandlers()
 	logger.Printf("waiting for async goroutines to finish...")
 	httpapi.WaitForAsync()
+	tracer.Stop()
```
bkeroackdsc: This is very minor, but I would defer this and move it up to the line after `startDatadogTracer()`.

pkg/nitro/env/env.go
```diff
@@ -98,6 +96,14 @@ func (m *Manager) pushNotification(ctx context.Context, env *newEnv, event notif
 		m.log(ctx, "pushNotification: %v: newenv.env is nil", event.Key())
 		return
 	}
+	span, _ := tracer.SpanFromContext(context.Background())
+	parentSpan, ok := GetSpanFromContext(ctx)
+	if !ok {
+		m.log(ctx, "pushNotification: error getting span from context")
+	} else {
+		span = tracer.StartSpan("process_env_config", tracer.ChildOf(parentSpan.Context()))
```
bkeroackdsc: I don't think this belongs here. `pushNotification()` just does a single notification, which is a relatively short operation.

pkg/nitro/env/env.go
```diff
@@ -150,14 +156,20 @@ func (m *Manager) pushNotification(ctx context.Context, env *newEnv, event notif
 	}
 }
 
-func (m *Manager) createPendingGithubStatus(ctx context.Context, rd *models.RepoRevisionData) error {
+func (m *Manager) createPendingGithubStatus(ctx context.Context, rd *models.RepoRevisionData) (err error) {
+	parentSpan, ok := GetSpanFromContext(ctx)
+	if !ok {
+		return errors.New("error generating new env: unable to get span from context")
+	}
+	span := tracer.StartSpan("create_pending_github_status", tracer.ChildOf(parentSpan.Context()))
```
bkeroackdsc: I think a better pattern to use is to write a new utility function that encapsulates the span retrieval and start:

```go
finish := startChildSpan(ctx, "create_pending_github_status")
defer finish()
// this could be done as a one-liner but it's much less readable:
// defer startChildSpan(ctx, "create_pending_github_status")()
```

If the context object doesn't contain a parent span, we can either just ignore and make the call a no-op, or start a new span without a parent. We shouldn't return an error, though, since errors in APM instrumentation should not cause us to abort environment operations.

pkg/nitro/env/env.go
```diff
@@ -193,20 +205,18 @@ func (m *Manager) createSuccessGithubStatus(ctx context.Context, rd *models.Repo
...
-			txn.NoticeError(err)
-		}
-	}()
-	ctx, cf := context.WithCancel(newNRTxnContext(ctx, txn))
+func (m *Manager) lockingOperation(ctx context.Context, operationName, repo, pr string, f func(ctx context.Context) error) (err error) {
+	parentSpan, ok := GetSpanFromContext(ctx)
+	if !ok {
+		return errors.New("error performing locking operation: unable to get span from context")
+	}
+	span := tracer.StartSpan(operationName, tracer.ChildOf(parentSpan.Context()))
```
bkeroackdsc: Keep in mind that every environment operation occurs as a locking operation, so I'm not sure how much value we get out of this span.

cmd/server.go
mikeykhalil: Good idea, not sure why i didn't do this in the first place.

pkg/nitro/env/env.go
mikeykhalil: sounds good.

pkg/nitro/env/env.go
mikeykhalil: Agreed, I like that pattern more too. Updated.

pkg/nitro/env/env.go
mikeykhalil: Yeah, this was originally being used to basically encapsulate the various async actions that acyl does (by passing the operation name argument). Now that you pointed this out, it seemed a bit simpler just to manage the spans within the operation functions themselves (e.g. `update`, `delete`, and `create`). 

Let me know if you had something else in mind or don't like the update I made.

mikeykhalil:
Thanks for the review. In addition to those changes, I also updated `acyl.yml` to point to the master branch of `helm-charts` once again, though we will need to merge the PR in that repo around the same time in order to avoid any issues when we go to deploy.

mikeykhalil:
Made one more update to remove all of the functions I wrote that handled creating contexts with spans, and extracting spans from contexts. Apparently the datadog tracing package had this functionality already.

mikeykhalil:
I agree w/ the points you made in the most recent review. Will make a follow up PR to address these issues. Merging.

PR #20, opened 2019-03-14
datadog apm: add sqlx and kubernetes-client ddtrace integrations

mikeykhalil:
1st commit: Add tracing around postgres and kubernetes client
2nd commit: Update mock `DataLayer` 
3rd commit: Update tests to work with this new mock
4th commit: Add tracing around consul (just lock and release for now)

edit: here are the commits I added after initial code review. 
5th commit: Replace spans with context in the Datalayer Interface, update implementations accordingly
6th commit: Fix cases where we were passing the incorrect context to the datalayer
7th commit: Add const for consul service name
8th commit: Move `wrapTransport` function into metahelm file (didn't really need it's own file)
9th commit: update furan client to allow easier mechanism for configuring datadog service name
10th commit: properly configure consul service name


mikeykhalil:
> I don't think that DataLayer methods should take spans directly. Rather they should take contexts, which may or may not contain spans. If present, they can be used, if a span is not present we don't trace in that package.


I was originally thinking the same thing, but I changed my mind after I implemented it that way. 

My reasoning was:
1. The user of DataLayer probably would not expect for DataLayer to basically ignore the context, grab a span from it and create a new context. The user would probably expect that the DataLayer would cancel its operations if they canceled that context.
2. On principle, it's probably best only to pass what is needed to a method. Since we don't really need context for anything else but a span, it seemed to make sense just to pass the span.
3. When we call `tracer.SpanFromContext(context.Background())`, this will actually create a no-op span and will not actually trace anything. So it seems like users have a mechanism to avoid tracing the DataLayer operations if they do so choose. 

However, I could understand that passing a context is more flexible in general if we decide to extend its usage. So I am not totally against changing it back to context.

bkeroackdsc:
> 1. The user of DataLayer probably would not expect for DataLayer to basically ignore the context, grab a span from it and create a new context.

DataLayer is an abstract interface describing methods on a generic data store. To me, spans feel like implementation details of one specific data backend that DataLayer may (or may not) be using. This would be analagous to passing in `sql.Txn` to a method, for example.

> The user would probably expect that the DataLayer would cancel its operations if they canceled that context.
 
This is a good point. We don't really have any explicitly long-running DataLayer methods, but to address this we might want to switch to using `sqlx` ...WithContext methods. Similarly we could check for a cancelled context at the top of each method and return an error if so.

> 2. On principle, it's probably best only to pass what is needed to a method. Since we don't really need context for anything else but a span, it seemed to make sense just to pass the span.

I agree with that principle usually, but in this case I think passing a context as the first function parameter is an established convention and keeps the method signatures generic and readable.

> 3. When we call `tracer.SpanFromContext(context.Background())`, this will actually create a no-op span and will not actually trace anything. So it seems like users have a mechanism to avoid tracing the DataLayer operations if they do so choose.

Only if the user reads the implementation code, in addition to the documentation for the datadog tracer package. It wasn't clear to me at first that this was the behavior for empty contexts.

> However, I could understand that passing a context is more flexible in general if we decide to extend its usage. So I am not totally against changing it back to context.

You make some valid points but my opinion is unchanged, I think contexts are the better choice.

mikeykhalil:
Fair points. I agree that it does feel like an implementation detail that ultimately shouldn't be a part of the interface. Will try to find some time to make thee updates today or on monday. thanks for the review and the explanations!

pkg/api/api_v0.go
```diff
@@ -323,34 +323,28 @@ func (api *v0api) envDetailHandler(w http.ResponseWriter, r *http.Request) {
 }
 
 func (api *v0api) envDestroyHandler(w http.ResponseWriter, r *http.Request) {
-	rootSpan := tracer.StartSpan("env_destroy_handler")
```
mikeykhalil: We already start spans for all http request handlers via `muxtrace`, so there is no need to start a rootspan here. We can grab the span from the request context and treat that as the root span.

mikeykhalil:
Made some updates based off the initial code review and cleaned up some other things I caught. 

Apologies for the huge PR. Tried to split it up over a bunch of different commits. 

Think it's ready for another review, when you have a chance, @bkeroackdsc. 

Thanks!

mikeykhalil:
Just realized i forgot to update some of the error value messages when detecting the context cancellations 🤦‍♂️ . Will fix those on Monday.

pkg/nitro/metahelm/metahelm.go
```diff
@@ -208,15 +210,25 @@ func NewInClusterK8sClientset(_, _ string) (*kubernetes.Clientset, *rest.Config,
...
 	if err != nil {
 		return nil, nil, errors.Wrap(err, "error getting k8s clientset")
 	}
 	return kc, kcfg, nil
 }
 
+// wrapTransport encapsulates the kubernetestrace WrapTransport and Kubernetes'
+// default TokenSource WrapTransport.
+func wrapTransport() func(rt http.RoundTripper) http.RoundTripper {
+	tokenFile := "/var/run/secrets/kubernetes.io/serviceaccount/token"
```
bkeroackdsc: This is already parameterized here: https://github.com/dollarshaveclub/acyl/blob/master/cmd/root.go#L34

It would be good if we could reuse that parameter in some way.

pkg/nitro/metahelm/metahelm.go
```diff
@@ -208,15 +210,25 @@ func NewInClusterK8sClientset(_, _ string) (*kubernetes.Clientset, *rest.Config,
 		return nil, nil, errors.Wrap(err, "error getting k8s in-cluster config")
 	}
 
-	// Since we are overriding WrapTranport, we need to set the token...
-	kcfg.WrapTransport = DatadogTraceWrapTransport()
+	kcfg.WrapTransport = wrapTransport()
```
bkeroackdsc: We should provide a way for the library user to opt out of tracing. For example, if a parameter isn't set (zero value), do not wrap the transport, or something similar.

mikeykhalil: That sounds good to me. Do we want to provide the ability to opt out of tracing from any of the libraries within this codebase? Or do we just want this in the context of the metahelm k8s client.

pkg/nitro/metahelm/metahelm.go
mikeykhalil: Good catch, will update. Didn't realize that we made this value configurable. 

`rest.InClusterConfig()` actually hard codes this token file path so we will probably still want to provide our own wrap transport even if the user of this library disables tracing. We just wont also include the `kubernetestrace.WrapRoundTripper` in that case.

pkg/nitro/metahelm/metahelm.go
bkeroackdsc: It's more of a general guideline. We should design packages (even internal ones like this) as if they would be consumed by the outside world. That means making behavior obvious, documented and configurable when possible.

mikeykhalil: That's fair. I made the tracing behavior configurable for the consul, kubernetes, and postgres clients.

PR #23, opened 2019-04-10
istio: add istio-injection label to nitro created namespaces

mikeykhalil:
Edit: Updated after initial review
This patch now does the following:

1.  Allows operators to specify labels at the CLI level which will be attached to all kubernetes resources that Nitro creates
2. Now, the cleanup process will basically only delete resources with ALL of the labels specified by the operator. Now that this is configurable, I tried to make it clear that it's important that these values are unique in the cluster so that Acyl doesn't cleanup resources that it didn't actually create.
3. It enables istio injection by default. Given our current istio installation, this doesn't inject sidecars automatically. It's still up to the charts to add annotations to a given pod template spec to enable automatic injection.

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -1,0 +1,58 @@
+package config
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestProcessLabels(t *testing.T) {
+	tests := []struct {
+		name           string
+		labelsStr      string
+		expectedErr    bool
+		expectedResult map[string]string
+	}{
+		{
+			name:        "single valid label",
+			labelsStr:   "acyl.dev/managed-by=nitro",
+			expectedErr: false,
+			expectedResult: map[string]string{
+				"acyl.dev/managed-by": "nitro",
+			},
+		},
+		{
+			name:        "multiple valid labels",
+			labelsStr:   "acyl.dev/managed-by=nitro,istio-injection=enabled",
+			expectedErr: false,
+			expectedResult: map[string]string{
+				"acyl.dev/managed-by": "nitro",
+				"istio-injection":     "enabled",
+			},
+		},
+		{
+			name:           "no label",
+			labelsStr:      "",
+			expectedErr:    true,
+			expectedResult: map[string]string{},
+		},
+		{
+			name:           "invalid labels",
+			labelsStr:      "potato,onion",
+			expectedErr:    true,
+			expectedResult: map[string]string{},
+		},
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			var k8scfg K8sConfig
+			err := k8scfg.ProcessLabels(tc.labelsStr)
+			receivedErr := err != nil
+			if receivedErr != tc.expectedErr {
+				t.Fatalf("K8sConfig.ProcessLabels received unexpected error: %v", err)
+			}
+			if !reflect.DeepEqual(k8scfg.Labels, tc.expectedResult) {
+				t.Fatalf("K8sConfig.ProcessLabels received unexpected results: %v", k8scfg.Labels)
+			}
+		})
+	}
+}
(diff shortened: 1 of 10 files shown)


pkg/nitro/metahelm/metahelm.go
```diff
@@ -640,6 +640,14 @@ func (ci ChartInstaller) GenerateCharts(ctx context.Context, ns string, newenv *
...
 	objLabelKey   = "acyl.dev/managed-by"
 	objLabelValue = "nitro"
+
+	// In order to allow automatic sidecar injection, we must label the
+	// namespace so that the istio MutatingWebhook can detect changes to pods
+	// within the namespace. We set the default value to disabled, but users
+	// can enable it by adding the annotation 'sidecar.istio.io/inject: "true"'
+	// to pods within the namespace.
+	istioInjectionLabel = "istio-injection"
+	istioInjectionValue = "disabled"
```
bkeroackdsc: If we're doing multiple labels, this should be a global `map[string]string` instead of multiple constants.

pkg/nitro/metahelm/metahelm.go
```diff
@@ -655,7 +663,8 @@ func (ci ChartInstaller) createNamespace(ctx context.Context, envname string) (s
 	ns := corev1.Namespace{
 		ObjectMeta: meta.ObjectMeta{
 			Labels: map[string]string{
-				objLabelKey: objLabelValue,
+				objLabelKey:         objLabelValue,
+				istioInjectionLabel: istioInjectionValue,
```
bkeroackdsc: Use the global map here instead.

pkg/nitro/metahelm/metahelm.go
bkeroackdsc: You might think about making this behavior toggle-able at the package and CLI level. Instead of hard-coding Istio-specific labels, allow the administrator to specify key-value labels to add to all created resources.

mikeykhalil: agreed about the global map. 

making this toggle-able sounds nice. the only caveat is that we technically just want these labels applied to the namespaces themselves, not necessarily all resources. so maybe the  the flag could be something like `--namespace-labels`.

bkeroackdsc: We only label resources created directly by Acyl (not resources created by charts), so it IIRC that would be namespaces and, optionally, ClusterRoleBindings for privileged repos.

pkg/config/config_test.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			var k8scfg K8sConfig
+			err := k8scfg.ProcessLabels(tc.labelsStr)
+			receivedErr := err != nil
+			if receivedErr != tc.expectedErr {
+				t.Fatalf("K8sConfig.ProcessLabels received unexpected error: %v", err)
+			}
+			if !reflect.DeepEqual(k8scfg.Labels, tc.expectedResult) {
```
bkeroackdsc: I suspect you used a table-driven test generator which uses `DeepEqual()` automatically, but I discourage DeepEqual in tests as it is often a cause of false test failures.

pkg/nitro/metahelm/metahelm.go
```diff
@@ -176,6 +179,7 @@ func NewChartInstallerWithClientsetFromContext(ib images.Builder, dl persistence
 		k8sgroupbindings: k8sGroupBindings,
 		k8srepowhitelist: k8sRepoWhitelist,
 		k8ssecretinjs:    k8sSecretInjs,
+		k8slabels:        k8sLabels,
```
bkeroackdsc: Check that `k8sLabels` isn't nil.

pkg/nitro/metahelm/metahelm.go
```diff
@@ -1059,12 +1054,24 @@ func (ci ChartInstaller) removeOrphanedNamespaces(ctx context.Context, maxAge ti
 	return nil
 }
 
+func (ci ChartInstaller) getLabelSelector() string {
```
bkeroackdsc: IIRC, there's a utility function in client-go that does this for you.

mikeykhalil:
Thanks for the review! Will get around to resolving them soon.

PR #26, opened 2019-05-16
WIP: Remove cleanup job

mikeykhalil:
Removes the cleanup cronjob since that isn't really needed and probably can do more harm than good. 

Also adds new flags for `cleanup` cmd to help operators that run this command locally.
1. `deletionDelay`
2. `dry-run`

These flags create ways for operators to be a bit more confident when going to cleanup stale resources. 

TODO: Probably want a flag to look at local kubecfg (like we do for tests), to make things easier for running this command locally.


bkeroackdsc:
Can we disable the cronjob by default, instead of deleting it entirely?

PR #27, opened 2019-06-04
GitHub commit status url template

mikeykhalil:
Some users were upset the Acyl V2 does not provide the QA environment's URL in the Github Commit Status. 

Given that Acyl V2 makes less assumptions about the environment and places the responsibility upon the helm charts to create ingress objects, it seems to make sense to just give users the ability to set the Target Url for a Github commit status. 

This PR allows users to create templates for setting the Github Commit Status for a given repo. This is done through the Acyl.yml. 

See the update Acyl.yml in this PR for example of how a user could configure this. 

This PR also added a new package, `go-cmp`. This is a safer alternative to `reflect.DeepEqual`. More details here: https://github.com/google/go-cmp

See https://github.com/dollarshaveclub/acyl/pull/28 for an example of this in use.

### DIFF
--- a/acyl.yml
+++ b/acyl.yml
@@ -28,13 +28,13 @@ dependencies:
     - chart_path: '.helm/charts/consul'
       chart_vars_path: './.helm/charts/consul/values.yaml'
 
-commit_statuses:
-  templates:
-    success:
-      description: "The Acyl environment {{ .EnvName }} was created successfully"
-      target_url: "https://{{ .EnvName }}-acyl.boogies.io"
-
 notifications:
+  github:
+    commit_statuses:
+      templates:
+        success:
+          description: "The Acyl environment {{ .EnvName }} was created successfully in the namespace {{ .K8sNamespace }}"
+          target_url: "https://{{ .EnvName }}-acyl.boogies.io"
   templates:
     success:
       title: "🏁 Success"
--- a/acyl.yml
+++ b/acyl.yml
@@ -29,12 +29,6 @@ dependencies:
       chart_vars_path: './.helm/charts/consul/values.yaml'
 
 notifications:
-  github:
-    commit_statuses:
-      templates:
-        success:
-          description: "The Acyl environment {{ .EnvName }} was created successfully in the namespace {{ .K8sNamespace }}"
-          target_url: "https://{{ .EnvName }}-acyl.boogies.io"
   templates:
     success:
       title: "🏁 Success"
--- a/acyl.yml
+++ b/acyl.yml
@@ -31,14 +31,8 @@ dependencies:
 commit_statuses:
   templates:
     success:
-      description: "Wow. Such success."
+      description: "The Acyl environment {{ .EnvName }} was created successfully"
       target_url: "https://{{ .EnvName }}-acyl.boogies.io"
-    pending:
-      description: "Wow. Much wait."
-      target_url: "https://giphy.com/gifs/happy-spongebob-squarepants-happiness-brHaCdJqCXijm"
-    failure:
-      description: "Wow. Much failure."
-      target_url: "https://giphy.com/gifs/windows-vaporwave-error-mq5y2jHRCAqMo"
 
 notifications:
   templates:
--- a/acyl.yml
+++ b/acyl.yml
@@ -28,6 +28,18 @@ dependencies:
     - chart_path: '.helm/charts/consul'
       chart_vars_path: './.helm/charts/consul/values.yaml'
 
+commit_statuses:
+  templates:
+    success:
+      description: "Wow. Such success."
+      target_url: "https://{{ .EnvName }}-acyl.boogies.io"
+    pending:
+      description: "Wow. Much wait."
+      target_url: "https://giphy.com/gifs/happy-spongebob-squarepants-happiness-brHaCdJqCXijm"
+    failure:
+      description: "Wow. Much failure."
+      target_url: "https://giphy.com/gifs/windows-vaporwave-error-mq5y2jHRCAqMo"
+
 notifications:
   templates:
     success:
--- a/pkg/models/commit_status.go
+++ b/pkg/models/commit_status.go
@@ -1,0 +1,102 @@
+package models
+
+// TODO (mk): Write tests, build and test locally, put up PR.
+
+import (
+	"bytes"
+	"text/template"
+
+	"github.com/pkg/errors"
+)
+
+type NitroCommitStatus int
+
+// Valid commit statuses from Nitro's perspective
+const (
+	// CommitStatusSuccess occurs when a Nitro environment has been created.
+	CommitStatusSuccess NitroCommitStatus = iota
+	// CommitStatusPending occurs when Nitro is first creating an environment,
+	// or updating an existing environment
+	CommitStatusPending
+	// CommitStatusFailure occurs when an unrecoverable error has occured during
+	// the creation or update of an environment
+	CommitStatusFailure
+)
+
+func (ncs NitroCommitStatus) Key() string {
+	switch ncs {
+	case CommitStatusSuccess:
+		return "success"
+	case CommitStatusPending:
+		return "pending"
+	case CommitStatusFailure:
+		return "failure"
+	default:
+		return "<unknown>"
+	}
+}
+
+// CommitStatuses models the configuration that Nitro supports for setting
+// commit statuses. Users can specify templates for each valid commit status.
+type CommitStatuses struct {
+	Templates map[string]CommitStatusTemplate `yaml:"templates" json:"templates"`
+}
+
+type CommitStatusTemplate struct {
+	Description string `yaml:"description" json:"description"`
+	TargetURL   string `yaml:"target_url" json:"target_url"`
+}
+
+type RenderedCommitStatus struct {
+	Description, TargetURL string
+}
+
+// CommitStatusData models the data available to commit status templates.
+type CommitStatusData struct {
+	EnvName string
+}
+
+// Render renders the commit status template using the supplied data.
+func (cs CommitStatusTemplate) Render(d CommitStatusData) (*RenderedCommitStatus, error) {
+	desc, err := renderTemplate("description", cs.Description, d)
+	if err != nil {
+		return nil, errors.Wrap(err, "error rendering commit status description template")
+	}
+	url, err := renderTemplate("target_url", cs.TargetURL, d)
+	if err != nil {
+		return nil, errors.Wrap(err, "error rendering commit status target url")
+	}
+	return &RenderedCommitStatus{
+		Description: desc,
+		TargetURL:   url,
+	}, nil
+}
+
+func renderTemplate(name, ts string, d CommitStatusData) (string, error) {
+	tmpl, err := template.New(name).Parse(ts)
+	if err != nil {
+		return "", errors.Wrap(err, "error parsing template")
+	}
+	res := &bytes.Buffer{}
+	if err := tmpl.Execute(res, d); err != nil {
+		return "", errors.Wrap(err, "error executing template")
+	}
+	return res.String(), nil
+}
+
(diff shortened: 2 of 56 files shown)


pkg/models/commit_status.go
```diff
@@ -0,0 +1,100 @@
...
+var DefaultCommitStatusTemplates = map[string]CommitStatusTemplate{
+	"success": CommitStatusTemplate{
+		Description: "The Acyl environment {{ .EnvName }} was created successfully.",
+		TargetURL:   "https://media.giphy.com/media/SRO0ZwmImic0/giphy.gif",
+	},
+	"pending": CommitStatusTemplate{
+		Description: "The Acyl environment {{ .EnvName }} is being created.",
+		TargetURL:   "https://media.giphy.com/media/oiymhxu13VYEo/giphy.gif",
+	},
+	"failure": CommitStatusTemplate{
```
esterlinej: I would consider updating `failure` to `error`, and providing the error state if possible, or a message to check their slack notification, though the point of this is to provide information at the github status PR page itself.

pkg/nitro/env/env.go
```diff
@@ -449,13 +428,13 @@ func (m *Manager) create(ctx context.Context, rd *models.RepoRevisionData) (envn
 				m.log(ctx, "error setting environment status to failed: %v", err)
 			}
 			m.pushNotification(ctx, newenv, notifier.Failure, "error creating: "+err.Error())
-			m.createErrorGithubStatus(ctx, rd)
+			m.setGithubCommitStatus(ctx, rd, newenv, models.CommitStatusFailure)
```
esterlinej: as noted prior, consider changing `failure` to `error`

pkg/nitro/env/env.go
```diff
@@ -623,12 +602,12 @@ func (m *Manager) update(ctx context.Context, rd *models.RepoRevisionData) (envn
 				m.log(ctx, "error setting environment status to failed: %v", err)
 			}
 			m.pushNotification(ctx, ne, notifier.Failure, err.Error())
-			m.createErrorGithubStatus(ctx, rd)
+			m.setGithubCommitStatus(ctx, rd, ne, models.CommitStatusFailure)
```
esterlinej: as noted prior, consider changing `failure` to `error`

pkg/models/commit_status.go
```diff
@@ -0,0 +1,100 @@
...
+type CommitStatusTemplate struct {
+	Description string `yaml:"description" json:"description"`
+	TargetURL   string `yaml:"target_url" json:"target_url"`
+}
+
+type RenderedCommitStatus struct {
+	Description, TargetURL string
+}
+
+// CommitStatusData models the data available to commit status templates.
```
bkeroackdsc: You'll probably want to include more data than just the environment name. Consider including everything that we supply to notification templates.

pkg/models/commit_status.go
```diff
@@ -0,0 +1,100 @@
...
+		return nil, errors.Wrap(err, "error rendering commit status target url")
+	}
+	return &RenderedCommitStatus{
+		Description: desc,
+		TargetURL:   url,
+	}, nil
+}
+
+func renderTemplate(name, ts string, d CommitStatusData) (string, error) {
+	tmpl, err := template.New(name).Parse(ts)
```
bkeroackdsc: Pre-parse the template on initial load, rather than doing it every time you render the template.

acyl.yml
```diff
@@ -28,6 +28,12 @@ dependencies:
     - chart_path: '.helm/charts/consul'
       chart_vars_path: './.helm/charts/consul/values.yaml'
 
+commit_statuses:
```
bkeroackdsc: This feels like it should be under the Notifications section, rather than a new top-level section on its own.

esterlinej: Solid point

mikeykhalil: I'd argue that an Acyl commit status is a single, mutable object while notifications in this context is a list of immutable messages. So there is a bit of a distinction.

However, I agree that having a `commit_status` as a top-level section could be distracting since it's an optional field that doesn't really impact the behavior of Acyl very much.

---

If we do want it under Notifications, would we want the structure to be:
```
notifications:
   github:
      commit_statuses:
```
or
```
notifications:
   commit_statuses:
```

I think we have a `notification.github` key already defined in the spec, though it appears to be only partially implemented as far as I can tell.

pkg/models/commit_status.go
bkeroackdsc: IIRC we use the term "failure" throughout the code and docs.

acyl.yml
bkeroackdsc: > I think we have a notification.github key already defined in the spec

Good catch. That would be a good place for `commit_statuses`.

pkg/models/commit_status.go
```diff
@@ -0,0 +1,100 @@
...
+func (ncs NitroCommitStatus) Key() string {
+	switch ncs {
+	case CommitStatusSuccess:
+		return "success"
+	case CommitStatusPending:
+		return "pending"
+	case CommitStatusFailure:
+		return "failure"
+	default:
+		return "<unknown>"
```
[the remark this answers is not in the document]
mikeykhalil: Good point. Updated to default to a failure.

pkg/models/commit_status.go
mikeykhalil: Good idea. I added `ErrorMessage` and `K8sNamespace`. The rest of the data was directly related to Github, so it seemed unnecessary to allow users to specify commit, PR, and Repo data within the commit status. If we think any other data could be valuable, i can pull the remaining variables that we supply to notification templates.

pkg/models/commit_status.go
mikeykhalil: I am fine with either `error` or `failure`, personally. I just copied what we had with the Notification templates. I left it as `failure` for now but I am happy to change it if there is reason to believe one is unclear.

On a related note, now there is an `ErrorMessage` value that can be used within the templates, which I think resolves part of your concern @esterlinej. So now users can have some information about the error that caused the failure.

pkg/models/commit_status.go
mikeykhalil: What do you mean by "initial load" here? 

We could potentially parse all 3 templates when we create or update the environment, and then pass these to the appropriate functions.

bkeroackdsc: You're right, withdrawn. I forgot that we need to parse the templates on every action because the user could've changed acyl.yml.

pkg/nitro/env/env.go
```diff
@@ -150,50 +141,47 @@ func (m *Manager) pushNotification(ctx context.Context, env *newEnv, event notif
...
 
-func (m *Manager) createErrorGithubStatus(ctx context.Context, rd *models.RepoRevisionData) error {
-	cs := &ghclient.CommitStatus{
-		Context:     "Acyl",
-		Status:      "failure",
-		Description: "Error creating environment",
-		TargetURL:   "https://media.giphy.com/media/pyFsc5uv5WPXN9Ocki/giphy.gif",
+func (m *Manager) setGithubCommitStatus(ctx context.Context, rd *models.RepoRevisionData, env *newEnv, ncs models.NitroCommitStatus, errmsg string) (*ghclient.CommitStatus, error) {
+	var csTemplate models.CommitStatusTemplate
+	if userProvidedTemplate, ok := env.rc.Notifications.GitHub.CommitStatuses.Templates[ncs.Key()]; ok {
```
bkeroackdsc:
```go
cst, ok := env.rc.Notifications.GitHub.CommitStatuses.Templates[ncs.Key()]
if !ok {
   cst = models.DefaultCommitStatusTemplates[ncs.Key()]
}
```

pkg/nitro/env/env.go
```diff
@@ -448,14 +436,15 @@ func (m *Manager) create(ctx context.Context, rd *models.RepoRevisionData) (envn
 			if err := m.DL.SetQAEnvironmentStatus(ctx, env.Name, models.Failure); err != nil {
 				m.log(ctx, "error setting environment status to failed: %v", err)
 			}
-			m.pushNotification(ctx, newenv, notifier.Failure, "error creating: "+err.Error())
-			m.createErrorGithubStatus(ctx, rd)
+			errmsg := "error creating: " + err.Error()
+			m.pushNotification(ctx, newenv, notifier.Failure, errmsg)
+			m.setGithubCommitStatus(ctx, rd, newenv, models.CommitStatusFailure, errmsg)
```
bkeroackdsc: Consider logging any error returned with `m.log()`. This will put it in the eventlog stream for debugging.

pkg/nitro/env/env.go
```diff
@@ -448,14 +436,15 @@ func (m *Manager) create(ctx context.Context, rd *models.RepoRevisionData) (envn
...
 			m.MC.Increment(mpfx+"create_errors", "triggering_repo:"+rd.Repo)
 			return
 		}
 		// metahelm.Manager sets the success status on QAEnvironment
 		m.pushNotification(ctx, newenv, notifier.Success, "")
-		m.createSuccessGithubStatus(ctx, rd)
+		m.setGithubCommitStatus(ctx, rd, newenv, models.CommitStatusSuccess, "")
```
bkeroackdsc: Log any error returned.

pkg/nitro/env/env.go
```diff
@@ -468,7 +457,7 @@ func (m *Manager) create(ctx context.Context, rd *models.RepoRevisionData) (envn
 		break
 	}
 	m.pushNotification(ctx, newenv, notifier.CreateEnvironment, "")
-	m.createPendingGithubStatus(ctx, rd)
+	m.setGithubCommitStatus(ctx, rd, newenv, models.CommitStatusPending, "")
```
bkeroackdsc: Log any error returned.

pkg/nitro/env/env.go
```diff
@@ -623,12 +612,12 @@ func (m *Manager) update(ctx context.Context, rd *models.RepoRevisionData) (envn
 				m.log(ctx, "error setting environment status to failed: %v", err)
 			}
 			m.pushNotification(ctx, ne, notifier.Failure, err.Error())
-			m.createErrorGithubStatus(ctx, rd)
+			m.setGithubCommitStatus(ctx, rd, ne, models.CommitStatusFailure, err.Error())
```
bkeroackdsc: Log any error returned.

pkg/nitro/env/env.go
```diff
@@ -623,12 +612,12 @@ func (m *Manager) update(ctx context.Context, rd *models.RepoRevisionData) (envn
...
 			return
 		}
 		// metahelm.Manager sets the success status on QAEnvironment
 		m.pushNotification(ctx, ne, notifier.Success, "")
-		m.createSuccessGithubStatus(ctx, rd)
+		m.setGithubCommitStatus(ctx, rd, ne, models.CommitStatusSuccess, "")
```
bkeroackdsc: Log any error returned.

pkg/nitro/env/env.go
```diff
@@ -648,7 +637,7 @@ func (m *Manager) update(ctx context.Context, rd *models.RepoRevisionData) (envn
 		break
 	}
 	m.pushNotification(ctx, ne, notifier.UpdateEnvironment, "")
-	m.createPendingGithubStatus(ctx, rd)
+	m.setGithubCommitStatus(ctx, rd, ne, models.CommitStatusPending, "")
```
bkeroackdsc: Log any error returned.

pkg/nitro/env/env.go
mikeykhalil: Good idea, added a `m.log().` statement in a deferred function within `setGithubCommitStatus`.

mikeykhalil:
Thanks for the reviews Ben and Joel! Think I addressed all of the concerns so far. Let me know if there is anything else we think should be changed.

esterlinej:
👍

PR #29, opened 2019-06-19
GitHub App support v1

### LINKED ISSUE #2: Convert Acyl to a GitHub App
If Acyl were a GitHub app (scoped only to the dollarshaveclub organization) it automatically creates an automated GUI for adding to a repository, so we wouldn't need Terraform to manage webhooks. In addition it would give us access to the new Checks API, which is a big improvement over commit statuses.

bkeroackdsc:
This is the first step to adding [GitHub App](https://developer.github.com/apps/) support.

Specifically, this implements a new webhook handler route (in parallel with our existing one which remains unchanged) that will receive authenticated GitHub App webhooks. All event handling that results from App webhooks will use the embedded request-scoped [GitHub credentials](https://developer.github.com/apps/building-github-apps/authenticating-with-github-apps/#authenticating-as-an-installation) (passed through via context) versus the static global GitHub API token used in the legacy webhook handler.

This will allow users to configure Acyl in their repos by [adding the App](https://developer.github.com/apps/installing-github-apps/) in the GitHub UI, instead of manually configuring webhooks. This also sets up the future ability for us to add a web UI with user login, and much richer [Checks API](https://developer.github.com/v3/checks/) integration (versus the current commit statuses).

This is not intended to change current functionality. We can gradually switch users/repos over to the GitHub App after merging and deployment.

Details:

- Event handling uses a new package: `pkg/ghapp`
- Support using a dynamic, per-request GitHub API client in `pkg/ghclient` using values passed in context
- Implement a new GH App webhook handler route
- Ignore events from forked HEAD repos (fixes security issue)
- Add a full integration test with a working GitHub App
- Add GitHub Delivery ID to eventlogs, for eventual lookup via Delivery IDs as well as event IDs
- Add support to the `webhook` cmd:
  - add a verbose flag
  - add required headers & fields

Fixes #2


Gopkg.toml
```diff
@@ -147,3 +147,8 @@
 [[constraint]]
   name = "gopkg.in/DataDog/dd-trace-go.v1"
   version = "1.10.0"
+
+[[constraint]]
+  name = "github.com/palantir/go-githubapp"
+  source = "github.com/dollarshaveclub/go-githubapp"
```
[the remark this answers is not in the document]
bkeroackdsc: https://github.com/palantir/go-githubapp/pull/13

pkg/ghapp/context/context.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+type GitHubClientContextKey string
+
+type githubAppClientFactory interface {
+	NewAppClient() (*github.Client, error)
+}
+
+// NewGitHubClientContext returns a context with the GitHub client from gha embedded as a value
+func NewGitHubClientContext(ctx stdlibctx.Context, gha githubAppClientFactory) (stdlibctx.Context, error) {
```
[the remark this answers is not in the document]
bkeroackdsc: Due to how GitHub apps work, we need to have a per-request GitHub client because authorization is done on a per-request basis. This allows one installation of a GitHub app to potentially service multiple organizations and repos, with each request having authorization scoped to only the org/repos required.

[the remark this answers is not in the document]
mikeykhalil: Got it. Initially, I was thinking that we could just pass the client as a separate argument. After seeing your PR for the upstream go-githubapp project, I now understand why we must use context here to pass the value.

pkg/api/api_v0.go
```diff
@@ -136,134 +160,164 @@ func setTagsForGithubWebhookHandler(span tracer.Span, rd *models.RepoRevisionDat
...
-			rootSpan.Finish()
 			log("success processing update event (env: %q); done", name)
 		}()
-	case ghevent.Destroy:
+	case "closed":
+		log("starting async processing for %v", action)
 		api.wg.Add(1)
 		go func() {
 			var err error
+			defer span.Finish(tracer.WithError(err))
```
mikeykhalil, line 224: Deferred call's arguments are evaluated immediately. So `err` might be `nil` at this point, but could be set later on in this function, i believe.

One way to potentially solve this issue would be:

``` 
defer func() {
     span.Finish(tracer.WithError(err))
}()
```
but we would need to confirm that the scope of `err`.

pkg/ghapp/pr.go
```diff
@@ -0,0 +1,124 @@
...
+	elogger, err := prh.getlogger(payload, did, rrd.Repo, rrd.PullRequest)
+	if err != nil {
+		return errors.Wrap(err, "error getting event logger")
+	}
+
+	// Create a new independent context for the async action
+	ctx = eventlogger.NewEventLoggerContext(ctx, elogger)
+
+	// Add the GitHub app client factory to the context
+	ctx = NewGitHubClientContext(ctx, event.GetInstallation().GetID(), prh)
```
mikeykhalil, line 98: So it seems we create a new Github Client for each event correct? Did a bit of digging and it seems like the go-githubapp caches the installation clients by default, but not the other clients. Perhaps we should consider caching a single or a small set of clients to be used to handle these requests, as a http.Client is safe for concurrent use by multiple goroutines.

Not a huge concern, as this current implementation is very convenient. Maybe just something to keep in mind while we are testing to see if any issues arise.

pkg/api/api_v0.go
bkeroackdsc: Good catch. Fixed.

pkg/ghapp/pr.go
bkeroackdsc: We currently only use installation clients (app clients only are used for app-related metadata). We can revisit if we need app clients in the future.

mikeykhalil: Ah, got it. Thanks for clearing that up 👍

PR #37, opened 2019-08-13
fix: use background context when seting qa environment status

mikeykhalil:
Fixes: https://github.com/dollarshaveclub/acyl/issues/39

When operations get canceled, we still want to be able to update the status of the qa environment through the data layer. So we should be passing the background context so that it can still update the status through the datalayer in these situations.

### DIFF
--- a/pkg/nitro/env/env.go
+++ b/pkg/nitro/env/env.go
@@ -187,7 +187,7 @@ func (m *Manager) setGithubCommitStatus(ctx context.Context, rd *models.RepoRevi
 		Description: renderedCSTemplate.Description,
 		TargetURL:   renderedCSTemplate.TargetURL,
 	}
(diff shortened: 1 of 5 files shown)


pkg/nitro/env/env.go
```diff
@@ -187,7 +187,7 @@ func (m *Manager) setGithubCommitStatus(ctx context.Context, rd *models.RepoRevi
 		Description: renderedCSTemplate.Description,
 		TargetURL:   renderedCSTemplate.TargetURL,
 	}
-	err = m.RC.SetStatus(ctx, rd.Repo, rd.SourceSHA, cs)
+	err = m.RC.SetStatus(context.Background(), rd.Repo, rd.SourceSHA, cs)
```
mikeykhalil: I believe this may also be the reason we have noticed issues where the github commit status never got updated...

PR #45, opened 2019-09-27
DQA traefik ingress

bkeroackdsc:
This switches Acyl DQAs to use Traefik ingress instead of Argo Tunnels.

### DIFF
--- a/acyl.yml
+++ b/acyl.yml
@@ -6,7 +6,7 @@ target_branches:
 
 application:
   chart_path: '.helm/charts/acyl'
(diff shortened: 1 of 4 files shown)


acyl.yml
```diff
@@ -6,7 +6,7 @@ target_branches:
 
 application:
   chart_path: '.helm/charts/acyl'
-  chart_vars_repo_path: 'dollarshaveclub/helm-charts@master:releases/kube-uw2-110/acyl/dqa.yml'
+  chart_vars_repo_path: 'dollarshaveclub/helm-charts@traefic-ingress:releases/kube-uw2-110/acyl/dqa.yml'
```
jmedberydsc: is this the right chart path?

PR #47, opened 2019-11-05
pkg/locker: replace perror with explicit error types

alrs:
This PR replaces `perror` with `PreemptedLockError` and `ContextCancelledError` so that the tests can verify correct errors with type assertions instead of `strings` comparison of `err.Error()`.

### DIFF
--- a/pkg/locker/preemptive_locker_test.go
+++ b/pkg/locker/preemptive_locker_test.go
@@ -3,7 +3,6 @@ package locker
 import (
 	"context"
 	"fmt"
-	"strings"
 	"testing"
 	"time"
 )
@@ -140,11 +139,12 @@ func TestPreemptiveLockerCancelledContext(t *testing.T) {
 
 	// should fail to acquire the lock because context is cancelled
 	_, err = ftl2.Lock(ctx2)
-	if err == nil {
-		t.Fatalf("should have returned a context cancelled error")
-	}
-	if !strings.Contains(err.Error(), "context was cancelled") {
-		t.Fatalf("unexpected error: %v", err)
+	switch err.(type) {
+	default:
+		t.Fatalf("wanted ContextCancelledError, got %q", err)
+	case *ContextCancelledError:
+		// success
+		t.Log(err)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


alrs:
@bkeroackdsc do you have any reservations about this kind of enhancement? Would you rather see the errors not exported outside the package?

bkeroackdsc:
Thanks @alrs ! Initially it seems OK to me but I'll look more in depth tomorrow and merge if it's ok.

pkg/locker/preemptive_locker_test.go
```diff
@@ -140,11 +139,12 @@ func TestPreemptiveLockerCancelledContext(t *testing.T) {
 
 	// should fail to acquire the lock because context is cancelled
 	_, err = ftl2.Lock(ctx2)
-	if err == nil {
-		t.Fatalf("should have returned a context cancelled error")
-	}
-	if !strings.Contains(err.Error(), "context was cancelled") {
-		t.Fatalf("unexpected error: %v", err)
+	switch err {
```
bkeroackdsc: This should be an explicit type switch.

pkg/locker/preemptive_locker_test.go
```diff
@@ -324,11 +324,12 @@ func TestPreemptiveLockerMultiPreemption(t *testing.T) {
 		// always get cancelled.
 		defer cf2()
 		_, err := ftl2.Lock(ctx2)
-		if err == nil {
-			t.Fatalf("second lock should have returned a preempted error")
-		}
-		if !strings.Contains(err.Error(), "preempted") {
-			t.Fatalf("unexpected error: %v", err)
+		switch err {
```
bkeroackdsc: Same here.

alrs:
@bkeroackdsc I'll most likely not be able to look at this until after Kubecon, are you guys down here this week?

bkeroackdsc:
@alrs @cleanshavenalex is there. Say hi!

pkg/locker/preemptive_locker_test.go
alrs: Done.

alrs:
I didn't find @cleanshavenalex, but the requested changes from @bkeroackdsc did make it in.

alrs:
I have a goroutine-in-tests fix lined up after this PR lands.

PR #50, opened 2019-11-15
nitro: more robust environment deletion

mikeykhalil:
This Patch ignores a namespace not found error when attempting to delete a namespace. Since there is nothing to delete, we do not need to return an error in this case.

### DIFF
--- a/pkg/nitro/metahelm/metahelm.go
+++ b/pkg/nitro/metahelm/metahelm.go
@@ -26,6 +26,7 @@ import (
 	billy "gopkg.in/src-d/go-billy.v4"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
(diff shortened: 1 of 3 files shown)


pkg/nitro/metahelm/metahelm.go
```diff
@@ -1010,8 +1014,11 @@ func (ci ChartInstaller) cleanUpNamespace(ctx context.Context, ns, envname strin
 	return nil
 }
 
-// DeleteNamespace deletes the kubernetes namespace and removes k8senv from the database
+// DeleteNamespace deletes the kubernetes namespace and removes k8senv from the database if they exist
 func (ci ChartInstaller) DeleteNamespace(ctx context.Context, k8senv *models.KubernetesEnvironment) error {
+	if k8senv == nil {
```
mikeykhalil: This part isn't really necessary to fix the corresponding issue. Just figured putting up a safe guard to avoid an unnecessary `panic` would be helpful.

bkeroackdsc: You probably should log something here, since a nil `k8senv` would be a clear application bug somewhere.

pkg/nitro/metahelm/metahelm.go
```diff
@@ -993,13 +994,16 @@ func (ci ChartInstaller) DeleteReleases(ctx context.Context, k8senv *models.Kube
...
 
+// cleanUpNamespace deletes an environment's namespace and ClusterRoleBinding, if they exist
 func (ci ChartInstaller) cleanUpNamespace(ctx context.Context, ns, envname string, privileged bool) error {
 	var zero int64
 	// Delete in background so that we can release the lock as soon as possible
 	bg := meta.DeletePropagationBackground
 	ci.log(ctx, "deleting namespace: %v", ns)
 	if err := ci.kc.CoreV1().Namespaces().Delete(ns, &meta.DeleteOptions{GracePeriodSeconds: &zero, PropagationPolicy: &bg}); err != nil {
-		return errors.Wrap(err, "error deleting namespace")
+		if !k8serrors.IsNotFound(err) {
```
bkeroackdsc: Add a comment explaining that we're swallowing up the error on purpose.

mikeykhalil: Good idea, done

bkeroackdsc:
@mikeykhalil Since you're possibly adding a new dependency package, be sure to run `dep ensure -v --dry-run` and then `dep ensure` to make sure vendored dependencies are in sync.

PR #55, opened 2019-12-11
pkg/locker: fix test goroutine to remove t.Fatal()

alrs:
`TestPreemptiveLockerMultiPreemption()` was using `testing.T.Fatal()` inside a goroutine. `T.Fatal()` is a convenience function that combines `T.Log()` and `T.FailNow()`. From the go doc for `T.FailNow()`:

> FailNow must be called from the goroutine running the test or benchmark function, not from other goroutines created during the test.

This changes the second test goroutine to exfiltrate errors via a channel, and use the `// allow time for the second process to contend` to check for an early exit for the second `Lock()`.

### DIFF
--- a/pkg/locker/preemptive_locker_test.go
+++ b/pkg/locker/preemptive_locker_test.go
@@ -316,8 +316,10 @@ func TestPreemptiveLockerMultiPreemption(t *testing.T) {
 	ftl2.w.watchInterval = 10 * time.Millisecond // low key watch interval
 
 	// Second process contends lock and will wait until preempted by process 3
+	errCh := make(chan error)
 	go func() {
 		defer close(process2)
+		defer close(errCh)
 		ctx2, cf2 := context.WithCancel(context.Background())
 		// we have to make sure the context gets cancelled, otherwise we leave a dangling goroutine
 		// contending the FakeRealPreemptableLock lock channel. In real life, the process context would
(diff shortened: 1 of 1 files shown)


alrs:
These test failures don't look to have anything to do with my changes, let me know if anything is needed from me.

alrs:
Standing by.

pkg/locker/preemptive_locker_test.go
```diff
@@ -326,15 +328,27 @@ func TestPreemptiveLockerMultiPreemption(t *testing.T) {
...
 	// allow time for the second process to contend
-	time.Sleep(10 * time.Millisecond)
+	select {
+	case err := <-errCh:
+		if err != nil {
+			// if an error comes down the channel
+			t.Fatalf("process 2 Lock(): %v", err)
+		} else {
+			// if a nil comes down the channel
+			t.Fatal("process 2 should not have returned yet")
```
bkeroackdsc: This feels brittle to me, since it's entirely based on how long the process 2 goroutine takes to exit. A nil error on `errChan` is expected because it will be closed when the goroutine exits (via defer).

Can you change this to not fail the test if a nil value is received on `errChan`?

alrs: Testing for a `nil` happens on line 380, once we're sure the goroutine should be done. If a nil shows up on 344, that means that goroutine #2 returned before goroutine #3 came to take the lock, which looks to be the point of the test.

bkeroackdsc:
I believe this is now obsolete due to the new PostgreSQL-based locking implementation.

PR #68, opened 2020-01-22
Timeout Notification

### LINKED ISSUE #58: Timeout error notifications not sent
It's not clear why or when, but sometimes when an operation times out no error notifications will be sent to the user. Instead the operation just appears to hang forever, and the last message in the event logs is:

```
"2020/01/09 18:33:18 event: xxxxxx: operation preempted: org/some-repo: 593\n"
```

bkeroackdsc:
Refactor operation timeouts to enable Acyl to send timeout notifications. Fixes #58.

- Enforce timeouts in `pkg/nitro/env` instead of simply letting the context cancel, so we can reliably send timeout notifications (previously they either were never sent, or with a cryptic "context cancelled" error message).
- (Delete operations are fully async so there is no meaningful timeout possible.)
- Increase the global operation timeout in the API layer as a safeguard against goroutines that hang forever.
- Add tests for timeouts and add some functionality to the datalayer fake.

### DIFF
--- a/.helm/charts/acyl/templates/deployment.yaml
+++ b/.helm/charts/acyl/templates/deployment.yaml
@@ -95,6 +95,10 @@ spec:
           - "{{ .Values.app.failure_reports.s3.keyprefix }}"
           - "--k8s-group-bindings"
           - "{{ .Values.app.k8s_group_bindings }}"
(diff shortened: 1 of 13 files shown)


esterlinej:
Helps when I select approve

PR #69, opened 2020-01-29
Remove helm deletes from environment updates

### LINKED ISSUE #59: Helm chart upgrade hangs
According to anecdotal reports, it appears that update-in-place operations (which utilize helm upgrade) appear to fail more frequently with timeouts and hangs in helm operations compared to create-from-scratch operations (helm install).

We should investigate the cause of this.

### LINKED ISSUE #61: Possible dangling goroutine
It's possible that we are leaving a dangling goroutine alive as I've noticed sometimes during deployments, old pods will hang waiting for all async actions to complete:

```2020/01/15 20:21:52 event: 53c8cbea-3f05-4d0a-94ec-e264a084adfa: event type: CommitPushEvent
2020/01/15 20:21:52 event: 53c8cbea-3f05-4d0a-94ec-e264a084adfa: fetching acyl.yml for dollarshaveclub/repo@0000000000000000000000000000000000000000
2020/01/15 20:21:52 event: 53c8cbea-3f05-4d0a-94ec-e264a084adfa: error fetching acyl.yml for webhook: error fetching acyl.yml: error getting GitHub repo contents: GET https://api.github.com/repos/dollarshaveclub/repo/contents/acyl.yml?ref=0000000000000000000000000000000000000000: 404 No commit found for the ref 0000000000000000000000000000000000000000 []
2020/01/15 20:21:52 {"error_details":"error fetching acyl.yml: error fetching acyl.yml: error getting GitHub repo contents: GET https://api.github.com/repos/dollarshaveclub/repo/contents/acyl.yml?ref=0000000000000000000000000000000000000000: 404 No commit found for the ref 0000000000000000000000000000000000000000 []"}
2020/01/15 20:21:52 webhook handler error: error fetching acyl.yml: error fetching acyl.yml: error getting GitHub repo contents: GET https://api.github.com/repos/dollarshaveclub/repo/contents/acyl.yml?ref=0000000000000000000000000000000000000000: 404 No commit found for the ref 0000000000000000000000000000000000000000 []
2020/01/15 20:21:53 event: 9a53e892-3be7-41f9-b2b0-5d66da21ac86: event type: PullRequestEvent
2020/01/15 20:21:53 event: 9a53e892-3be7-41f9-b2b0-5d66da21ac86: fetching acyl.yml for dollarshaveclub/repo@xxx
2020/01/15 20:21:53 event: 9a53e892-3be7-41f9-b2b0-5d66da21ac86: PR base branch not found in target branches: master
2020/01/15 20:21:53 event: 9a53e892-3be7-41f9-b2b0-5d66da21ac86: github event: relevancy check failed: Update: {synchronize {repo dollarshaveclub/repo} {17093 {user} {release xxx} {master xxx}}  xxx xxx {user}}
2020/01/15 20:21:53 event: 9a53e892-3be7-41f9-b2b0-5d66da21ac86: event not relevant: NotRelevant
2020/01/15 20:25:08 http: Server closed
2020/01/15 20:25:08 waiting for handlers to finish...
2020/01/15 20:25:08 waiting for async goroutines to finish...```

### PR DESCRIPTION
**Pull request by bkeroackdsc**
This changes update behavior to attempt to address issues with helm delete.

Previously, if a config signature mismatch was detected during an environment update, we would attempt to helm delete all releases (basically, uninstall all k8s resources) and then re-use the namespace for a fresh set of helm installs. The rationale was to avoid recreating namespaces whenever possible (which includes installing Tiller), but it turns out that helm deletes frequently hang (#59). 

They also take much longer to perform than anticipated (5-10 seconds per release) so it's unlikely that we save any time versus deleting namespaces and re-provisioning from scratch.

Finally, since helm delete method calls do not accept contexts, if they hang we are potentially leaking goroutines (#61). We also can't set any timeouts on them.

For these reasons, this PR alters updates to delete and recreate namespaces when the config signature doesn't match. The primary user-visible change here is that the namespace name may change when users request an environment update (via PR commit push). We don't want to re-use the same namespace name because then we'd have to wait for k8s to fully finish deleting the old namespace, and that can take a while with finalizers, etc.

Also, unrelated to the functional changes, this PR includes a small utility for populating a local Acyl DQA (minikube, microk8s) with necessary secrets for full operational use: https://github.com/dollarshaveclub/acyl/pull/69/commits/16b5d6abb1dff7bfa171e037aa204b7ec7d8c2ea

Fixes #59. Fixes #61 .

### CODE REVIEW

File: .helm/charts/acyl/templates/deployment.yaml [hunk 1]
```diff
@@ -168,6 +168,11 @@ spec:
               secretKeyRef:
                 name: dummy-acyl-secrets
                 key: db_uri
+          - name: ACYL_K8S_IMAGE_PULL_SECRET
```
mikeykhalil: Doesn't look like we use this environment variable, unless I am missing something? Seems like the `localsecrets` utility just reads from a file.

bkeroackdsc: It's a little bit obscure, but that block of env vars is used when `secrets_from_env` is defined, which uses the [pvc environment variable backend](https://godoc.org/github.com/dollarshaveclub/pvc#WithEnvVarBackend). So in this case the secret that would be in vault at `<prefix>/k8s/image_pull_secret` is instead at this computed env var name.

PR #78, opened 2020-04-07
Acyl Cleanup, Removing Amino Support

esterlinej:
**Overview**
Removes support for building Acyl v1 Amino environments, while supporting the interface necessary for Acyl v2 Nitro environments, and refactoring key features necessary to support building multiple environment versions.

**Key Changes**
- **cmd/server:** Removed unused variables and flags. Refactored `server()` command to support Acyl v2 Nitro environments only; nitro manager.
- **pkg/config:** Removed not needed `AminoConfig` struct and supporting `Parse()` command.
- **pkg/mocks:** Removed deprecated package.
- **pkg/nitro:** Removed not needed files, `env/combined_spawner.go` and `env/combined_spawner_test.go`.
- **pkg/spawner:** Removed `pbaminoapi/pbaminoapi.pb.go`, `spawner_test.go`, `amino_backend.go`, `backends.go`, and `aliases.go`. Majority of `spawner.go` has been removed, with the `EnvironmentSpawner` interface.

**v1 Amino Error Handling**
```
error creating: error processing environment config: error validating environment config: 
error getting metadata: error processing target repo acyl.yml: acyl.yml is previous or 
unsupported version
```
**v2 Nitro Success Message**
```
fanciful-coordinate-system
dollarshaveclub/acyl PR #1 amino-cleanup
K8s Namespace: nitro-45591-fanciful-coordinate-system
https://fanciful-coordinate-system-acyl.my.domain
```

**Manual Testing**
Multiple successful builds; Acyl v2 Nitro environment for the acyl app, then used that environment to successfully build another Acyl v2 Nitro environment for the acyl app.


pkg/spawner/aliases.go
```diff
@@ -1,72 +1,9 @@
 package spawner
```
bkeroackdsc, line 1: I would get rid of all of these type aliases. They were only useful back when packages were split out a long time ago.

pkg/spawner/spawner_test.go
```diff
@@ -1,1155 +0,0 @@
...
-	"context"
-	"fmt"
-	"io/ioutil"
-	"log"
-	"strings"
-	"testing"
-	"time"
-
-	"github.com/dollarshaveclub/acyl/pkg/ghclient"
-	"github.com/dollarshaveclub/acyl/pkg/mocks"
```
bkeroackdsc, line 13: I believe (IIRC) that these tests are the only use of the autogenerated mocks in `pkg/mocks` so we can probably get rid of that entire package as well.

esterlinej: Some of these are mocks for other packages, but I will remove those specific to `pkg/spawner`.

PR #84, opened 2020-04-29
GitHub App Auth for UI routes

bkeroackdsc:
This implements an optional OAuth authentication and authorization flow for UI web pages, based upon GitHub App settings. You must have a GitHub app set up and the running Acyl instance must be configured with the relevant settings. We maintain a 1:1 mapping of GitHub app installation to running Acyl instance (Acyl is not multitenant for app installations currently).

* [High-level overview of auth flow](https://github.com/dollarshaveclub/acyl/blob/ui-login/pkg/api/api_ui.go#L274-L306)

A GitHub user is considered authorized if and only if all the following are true:
1. the user has explicit permissions for this installation of this GitHub app
1. the user has at least one repository accessible via this installation

Most of the code is in files that the PR diff view does not render by default:

* https://github.com/dollarshaveclub/acyl/pull/84/files#diff-91c8bd50b27311a1ffadda6a7b249f28
* (tests) https://github.com/dollarshaveclub/acyl/pull/84/files#diff-4629729115116ef5c8b50cea3b369f11


pkg/ghclient/apps.go
```diff
@@ -0,0 +1,101 @@
...
+// GetUserAppInstallationCount returns the number of app installations that are accessible to the authenticated user
+// This method only uses the static token associated with the GitHubClient and not anything present in the context
+// GitHubClient should be populated with the user token returned by the oauth login endpoint via the oauth callback handler
+func (ghc *GitHubClient) GetUserAppInstallations(ctx context.Context) (AppInstallations, error) {
+	lopt := &github.ListOptions{PerPage: 100}
+	out := []*github.Installation{}
+	for {
+		ctx, cf := context.WithTimeout(ctx, ghTimeout)
+		defer cf()
+		insts, resp, err := ghc.c.Apps.ListUserInstallations(ctx, lopt)
```
mikeykhalil, line 53: Think this endpoint is still in the preview period: 

https://developer.github.com/v3/apps/installations/#list-installations-for-a-user

Likely not a huge deal, but could be something we need to keep an eye on if the API ends up changing a bit.

PR #88, opened 2020-05-14
UI Enhancement

bkeroackdsc:
This adds a fuller UI experience for the GitHub app including more views and finer-grained permissions.

### Views

- **Home**: show all environments for the authenticated GitHub user, or all envs for all visible repos.
image
- **Environment**: show all events for an environment, with relevant metadata and links to each event in chronological order.
image
- **Event status**: same as before, but adds a breadcrumb display back to the other UI views.
image

### Permissions

- To view the home page, you must be an authenticated GitHub user with read access to at least one repo to which the GitHub app has access.
- To view the environment page, you must be an authenticated GitHub user with read access to the specific repo with which the environment is associated.
- To view the event status page, the same permissions are required as the home page (ie: authenticated GitHub user with read access to at least one repo). This is subject to change but is how the UI works prior to this PR.


pkg/api/middleware.go
```diff
@@ -106,6 +131,9 @@ func (sa *sessionAuthenticator) sessionAuth(f http.HandlerFunc) http.HandlerFunc
 	if !sa.Enforce {
 		return f
 	}
+	log := func(msg string, args ...interface{}) {
+		log.Printf("sessionAuth: "+msg, args...)
```
mikeykhalil, line 135: We lose some compile time errors when using dynamically generated format specifiers. 

For example, when using a static format specifier, you get some nice build time error when the number of formatting directives != the number of arguments. 

https://play.golang.org/p/KzJr_NY3-St

Maybe consider instantiating a new logger for this struct and setting the prefix for the logger, accordingly?

bkeroackdsc: Agreed. I cleaned up the dynamic log function and added a reusable routeLogger which prepends the request route to log messages.

PR #92, opened 2020-06-25
Feature/rebuild env

esterlinej:
It would be useful for users to be able to trigger rebuilds of existing environments in the UI, without having to manually open/close a PR or push a dummy commit.

We would like to implement an “actions” menu within the web UI on the environment details page, with items that trigger a rebuild (_see details_). Once triggered, a rebuild is an event exactly as if the user had opened the PR or pushed a new commit, and can be monitored in the UI in the same way.

### Details
A “rebuild” is simply a synthetic event created by Acyl and executed in exactly the same way as an event received by the GitHub app webhook.

There should be two rebuild options:

- **Fast Rebuild:** this creates a “synchronize” event with the same revision data as the last PR event. This will usually result in helm upgrades and a relatively quick rebuild processing time.
- **Full Rebuild:** this creates a “synchronize” event with a different config signature (_the implementation may just clear the config signature as a way of forcing a mismatch_) which results in a complete teardown of the existing environment and a rebuild from scratch. This will take longer but should guarantee that all resources are rebuilt.

---------------

**k8s 1.16.x requirements:** Update helm chart deployment for consul, image sha for postgresql


pkg/api/api_v2.go
```diff
@@ -710,3 +790,83 @@ func (api *v2api) userEnvDetailHandler(w http.ResponseWriter, r *http.Request) {
...
+		Repo:         qae.Repo,
+		PullRequest:  qae.PullRequest,
+		SourceSHA:    qae.SourceSHA,
+		BaseSHA:      qae.BaseSHA,
+		SourceBranch: qae.SourceBranch,
+		BaseBranch:   qae.BaseBranch,
+		SourceRef:    qae.SourceRef,
+		IsFork:       false, // TODO determine IsFork
+	}
+	if !fullRebuild {
```
bkeroackdsc, line 853: If a full rebuild is requested, the easiest way to [force a full rebuild](https://github.com/dollarshaveclub/acyl/blob/master/pkg/nitro/env/env.go#L757-L758) is to clear the [config signature](https://github.com/dollarshaveclub/acyl/blob/master/pkg/models/nitro.go#L249).

You'll need to implement a [new datalayer method](https://github.com/dollarshaveclub/acyl/blob/master/pkg/persistence/datalayer.go#L57-L63) to do this (`UpdateK8sEnvConfigSignature()`.

pkg/api/api_v2.go
```diff
@@ -710,3 +790,83 @@ func (api *v2api) userEnvDetailHandler(w http.ResponseWriter, r *http.Request) {
...
+		err = api.processWebhookV2(r.Context(), "synchronize", rrd)
+		if err != nil {
+			api.rlogger(r).Logf("finished processing update with error: %v", err)
+			return
+		}
+	} else {
+		err = api.processWebhookV2(r.Context(), "closed", rrd)
```
bkeroackdsc, line 860: Eliminate the calls to the separate handler and create a new goroutine that invokes `EnvironmentSpawner.Update()` in this handler instead.

pkg/api/api_v2.go
```diff
@@ -710,3 +790,83 @@ func (api *v2api) userEnvDetailHandler(w http.ResponseWriter, r *http.Request) {
...
+		PullRequest:  qae.PullRequest,
+		SourceSHA:    qae.SourceSHA,
+		BaseSHA:      qae.BaseSHA,
+		SourceBranch: qae.SourceBranch,
+		BaseBranch:   qae.BaseBranch,
+		SourceRef:    qae.SourceRef,
+		IsFork:       false, // TODO determine IsFork
+	}
+	if !fullRebuild {
+		err = api.processWebhookV2(r.Context(), "synchronize", rrd)
```
bkeroackdsc, line 854: Do not use the request context (`r.Context()`) since that gets cancelled when this TCP connection is closed. If you derive the context that's used by your async goroutine from the request context, it will get cancelled as soon as you return an HTTP response back to the client.

pkg/api/middleware.go
```diff
@@ -219,3 +219,25 @@ func (up *userPermissions) GetUserVisibleRepos(ctx context.Context, uis models.U
 	}
 	return out, nil
 }
+
+// GetUserWritableRepos returns the names of all repos (owner/repo) for which the authenticated user has "admin" or "push" permissions
+func (up *userPermissions) GetUserWritableRepos(ctx context.Context, uis models.UISession) ([]string, error) {
```
bkeroackdsc, line 224: Nice!

In retrospect, I think `userPermissions` methods should return maps of repos instead of slices (like `ghc.GetUserAppRepoPermissions()` returns) which will make it easier to check for the presence of a specific repo.

pkg/api/api_v2.go
```diff
@@ -710,3 +717,119 @@ func (api *v2api) userEnvDetailHandler(w http.ResponseWriter, r *http.Request) {
...
+	rrd := models.RepoRevisionData{
+		User:         qae.User,
+		Repo:         qae.Repo,
+		PullRequest:  qae.PullRequest,
+		SourceSHA:    qae.SourceSHA,
+		BaseSHA:      qae.BaseSHA,
+		SourceBranch: qae.SourceBranch,
+		BaseBranch:   qae.BaseBranch,
+		SourceRef:    qae.SourceRef,
+		IsFork:       false, // TODO determine IsFork
```
bkeroackdsc, line 777: This is definitely false. This line can be removed.

pkg/api/api_v2.go
```diff
@@ -710,3 +717,119 @@ func (api *v2api) userEnvDetailHandler(w http.ResponseWriter, r *http.Request) {
...
+	}
+	messaging := "rebuild"
+	if fullRebuild {
+		k8senv.ConfigSignature = []byte{}
```
bkeroackdsc, line 781: This is insufficient. You'll need to [change this value in the database](https://github.com/dollarshaveclub/acyl/blob/master/pkg/persistence/datalayer.go#L58-L62) since this value will get fetched by `nitro/env.Manager` during event processing.

Also note that the value needs to be exactly 32 bytes long. The zero value of an array should suffice:

```go
dl.UpdateK8sEnvConfSignature(name, [32]byte{})
```

pkg/api/api_v2.go
```diff
@@ -710,3 +717,119 @@ func (api *v2api) userEnvDetailHandler(w http.ResponseWriter, r *http.Request) {
...
+		messaging = "full rebuild"
+	}
+
+	b, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		api.badRequestError(w, err)
+		return
+	}
```
bkeroackdsc, line 789: The API call doesn't have anything in the request body so this is unneeded.

pkg/api/api_v2.go
```diff
@@ -710,3 +717,119 @@ func (api *v2api) userEnvDetailHandler(w http.ResponseWriter, r *http.Request) {
...
+	out, err := api.ge.New(b, did, sig)
+	if err != nil {
+		_, bsok := err.(ghevent.BadSignature)
+		if bsok {
+			api.forbiddenError(w, "invalid hub signature")
+		} else {
+			api.badRequestError(w, err)
+		}
+		return
+	}
```
bkeroackdsc, line 810: This is only relevant for requests received via the GitHub webhook handler and isn't needed here.

pkg/api/api_v2.go
```diff
@@ -710,3 +717,119 @@ func (api *v2api) userEnvDetailHandler(w http.ResponseWriter, r *http.Request) {
...
+	ctx := eventlogger.NewEventLoggerContext(context.Background(), out.Logger)
```
bkeroackdsc, line 811: You'll need to [create a new eventlogger](https://github.com/dollarshaveclub/acyl/blob/master/pkg/eventlogger/logger.go#L14-L44) first and then embed it in the cotext with this method.

Here's an example of how the eventlogger is created for GitHub app webhook events: https://github.com/dollarshaveclub/acyl/blob/master/pkg/ghapp/pr.go#L108-L121

pkg/api/middleware.go
esterlinej: Yeah, seeing that this was in use already, I opted to add a new function for Write/Admin. I didn't want to alter existing, and force changes down the line, but this thought had crossed my mind. That said, `rps, err := ghc.GetUserAppRepoPermissions(ctx, up.instID)` already returns the map, we could alter this function to simply return the map, and process the write/admin repos in the handler, providing expanded use for future iterations.

pkg/persistence/k8s_envs.go
```diff
@@ -65,6 +65,16 @@ func (pg *PGLayer) UpdateK8sEnvTillerAddr(ctx context.Context, envname, taddr st
 	return errors.Wrap(err, "error updating k8s environment")
 }
 
+// UpdateK8sEnvConfSignature updates an existing k8s environment config signature
+func (pg *PGLayer) UpdateK8sEnvConfSignature(ctx context.Context, name string, confSig [32]byte) error {
```
bkeroackdsc, line 69: 👍 This needs a test and a fake implementation.

pkg/api/middleware.go
bkeroackdsc: I would probably implement it with the existing methods, but return a `map[string]struct{}` (map of string to empty struct) instead.

That way you could easily check if a repo exists without having to iterate through the slice.

pkg/api/api_v2.go
```diff
@@ -710,3 +717,131 @@ func (api *v2api) userEnvDetailHandler(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		api.rlogger(r).Logf("error getting random UUID: %v", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	did, err := uuid.Parse(r.Header.Get("X-GitHub-Delivery"))
+	if err != nil {
+		// Ignore invalid or missing Delivery ID
+		did = uuid.Nil
+	}
```
bkeroackdsc, line 789: `X-GitHub-Delivery` definitely won't be there, and we wouldn't use it even if it was because this isn't a GitHub webhook handler.

It's OK to use `uuid.Nil` here. Technically we treat GitHub Delivery ID as [another unique lookup key](https://github.com/dollarshaveclub/acyl/blob/master/migrations/0016_event_logs_delivery_id_up.sql) but I think it's best to use the zero value here for synthetic events (vs a random one).

pkg/api/api_v2.go
```diff
@@ -710,3 +717,131 @@ func (api *v2api) userEnvDetailHandler(w http.ResponseWriter, r *http.Request) {
...
+	elogger := &eventlogger.Logger{
+		ID:         id,
+		DeliveryID: did,
+		DL:         api.dl,
+		Sink:       os.Stdout,
+	}
+	if err := elogger.Init(body, qae.Repo, qae.PullRequest); err != nil {
```
bkeroackdsc, line 796: `body` isn't set to anything so you might as well use `nil` here.

pkg/api/api_v2.go
```diff
@@ -710,3 +717,131 @@ func (api *v2api) userEnvDetailHandler(w http.ResponseWriter, r *http.Request) {
...
+		api.rlogger(r).Logf("error initializing event logger: %v", err)
+		w.WriteHeader(http.StatusBadRequest)
```
bkeroackdsc, line 798: Use 500 Internal Server Error for errors that aren't the client's fault.

pkg/api/api_v2.go
```diff
@@ -710,3 +717,131 @@ func (api *v2api) userEnvDetailHandler(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+	if err := elogger.SetEnvName(qae.Name); err != nil {
+		api.rlogger(r).Logf("error setting event logger name: %v", err)
+		w.WriteHeader(http.StatusBadRequest)
```
bkeroackdsc, line 803: Same here.

pkg/api/api_v2.go
```diff
@@ -710,3 +717,131 @@ func (api *v2api) userEnvDetailHandler(w http.ResponseWriter, r *http.Request) {
...
+		api.rlogger(r).Logf("error getting k8s env from db: %v", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	if k8senv == nil {
+		api.rlogger(r).Logf("k8s env not found")
+		w.WriteHeader(http.StatusNotFound)
+		return
+	}
+	rrd := models.RepoRevisionData{
```
bkeroackdsc, line 766: I forgot that there's already a method for this: https://github.com/dollarshaveclub/acyl/blob/master/pkg/models/models.go#L221

pkg/api/api_v2.go
```diff
@@ -710,3 +717,117 @@ func (api *v2api) userEnvDetailHandler(w http.ResponseWriter, r *http.Request) {
...
+		w.WriteHeader(http.StatusNotFound)
+		return
+	}
+	repos, err := userPermissionsClient(api.oauth).GetUserWritableRepos(r.Context(), uis)
+	if err != nil {
+		api.rlogger(r).Logf("error getting user writable repos: %v", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	if repos[qae.Repo].Repo == "" {
```
bkeroackdsc, line 751:
```go
if _, ok := repos[qae.Repo]; !ok {
  // ...
}
```

pkg/api/api_v2.go
```diff
@@ -710,3 +717,117 @@ func (api *v2api) userEnvDetailHandler(w http.ResponseWriter, r *http.Request) {
...
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	elogger := &eventlogger.Logger{
+		ID:         id,
+		DeliveryID: uuid.Nil,
+		DL:         api.dl,
+		Sink:       os.Stdout,
+	}
+	if err := elogger.Init(nil, qae.Repo, qae.PullRequest); err != nil {
```
bkeroackdsc, line 782: This is a nitpick, but I would use an empty, non-nil byteslice (`[]byte{}`) to signify an empty payload, instead of nil.

pkg/api/api_v2.go
```diff
@@ -710,3 +717,117 @@ func (api *v2api) userEnvDetailHandler(w http.ResponseWriter, r *http.Request) {
...
+	span := tracer.StartSpan("actions_rebuilder")
+	span.SetTag(ext.SamplingPriority, ext.PriorityUserKeep)
+	setTagsForGithubWebhookHandler(span, *rrd)
+	ctx = tracer.ContextWithSpan(ctx, span)
+	logger := eventlogger.GetLogger(ctx).Printf
+
+	// check for full rebuild
+	var fullRebuild bool
+	messaging := "rebuild"
+	fullRebuild = r.URL.Query().Get("full") == "true"
```
bkeroackdsc, line 805: You can eliminate the variable declaration above: `fullRebuild := r.URL.Query().Get("full") == "true"`

pkg/api/api_v2.go
```diff
@@ -710,3 +717,117 @@ func (api *v2api) userEnvDetailHandler(w http.ResponseWriter, r *http.Request) {
...
+	if fullRebuild {
```
bkeroackdsc, line 806: Since this is the only place where this boolean is used, you can eliminate it entirely:

```go
if r.URL.Query().Get("full") == "true" {
  // ...
}
```

pkg/api/middleware.go
```diff
@@ -219,3 +219,25 @@ func (up *userPermissions) GetUserVisibleRepos(ctx context.Context, uis models.U
...
+		return nil, errors.Wrap(err, "error decrypting user token")
+	}
+
+	ghc := up.gcfunc(tkn)
+
+	rps, err := ghc.GetUserAppRepoPermissions(ctx, up.instID)
+	if err != nil {
+		return nil, errors.Wrap(err, "error getting user repos")
+	}
+	out := map[string]ghclient.AppRepoPermissions{}
```
bkeroackdsc, line 236: Since you know the map size, you can initialize it like this which is more efficient: 
```go
out := make(map[string]ghclient.AppRepoPermissions, len(rps))
```

pkg/api/testdata
```diff
@@ -1 +0,0 @@
-../testdata/
```
bkeroackdsc, line 1: This symlink shouldn't be deleted.

pkg/persistence/datalayer.go
```diff
@@ -60,6 +60,7 @@ type K8sEnvDataLayer interface {
 	CreateK8sEnv(ctx context.Context, env *models.KubernetesEnvironment) error
 	DeleteK8sEnv(ctx context.Context, name string) error
 	UpdateK8sEnvTillerAddr(ctx context.Context, envname, taddr string) error
+	UpdateK8sEnvConfSignature(ctx context.Context, name string, confSig [32]byte) error
```
bkeroackdsc, line 63: Nitpick: I would name this `UpdateK8sEnvConfigSignature`.

pkg/persistence/fake.go
```diff
@@ -754,6 +754,21 @@ func (fdl *FakeDataLayer) UpdateK8sEnvTillerAddr(ctx context.Context, envname, t
...
+	if isCancelled(ctx) {
+		return ctx.Err()
+	}
+	fdl.doDelay()
+	fdl.data.Lock()
+	defer fdl.data.Unlock()
+	env, ok := fdl.data.k8s[name]
+	if ok {
+		env.ConfigSignature = confSig[:]
+		fdl.data.k8s[name] = env
```
bkeroackdsc, line 767: `env` is a pointer, so you can modify directly: `fdl.data.k8s[name].ConfigSignature = confSig[:]`

ui/views/env.html
```diff
@@ -39,6 +39,126 @@ <h2 class="mb-0">
...
+                                                            class="close"
+                                                            data-dismiss="modal"
+                                                            aria-label="Close"
+                                                    >
+                                                        <span aria-hidden="true">&times;</span>
+                                                    </button>
+                                                </div>
+                                                <div class="modal-body">
+                                                    <small>
+                                                        <p class="text-sm-left">Fully rebuild the current environment including dependencies.</p>
```
esterlinej, line 139: The modal description text isn't final, just added something until we determine what each should state.

ui/assets/env.js
```diff
@@ -166,3 +166,30 @@ document.addEventListener("DOMContentLoaded", function(){
...
+document.addEventListener("DOMContentLoaded", function() {
+    document.getElementById("actionsBtn").addEventListener('click', function () {
+        $('#synchronizeModalConfirm').on('click', function () {
+            rebuild();
+        });
+        $('#rebuildModalConfirm').on('click', function () {
+            rebuild(true);
+        });
+    });
+});
```
bkeroackdsc, line 195: Move this over to the existing `DOMContentLoaded` handler function.

ui/views/env.html
```diff
@@ -39,6 +39,126 @@ <h2 class="mb-0">
...
                                     </button>
                                     <button id="refreshbtn" class="fas btn btn-sm btn-secondary">&#xf2f1;</button>
+                                    <div id="actionsGroup" class="btn-group btn-group-sm">
+                                        <button
+                                                type="button"
+                                                id="actionsBtn"
+                                                class="btn btn-secondary dropdown-toggle"
+                                                data-toggle="dropdown"
+                                        >
+                                            <i class="fas fa-cog"></i>
```
bkeroackdsc, line 49: Utilize the same unicode literal method we use for the refresh button, so you don't need to import another javascript file (the FontAwesome js).

```html
<button id="refreshbtn" class="fas btn btn-sm btn-secondary">&#xf2f1;</button>
```

ui/views/partials/footer.html
```diff
@@ -23,4 +23,5 @@
             integrity="sha384-wfSDF2E50Y2D1uUdj0O3uMBJnjuUD4Ih7YwaYd1iqfktj0Uod8GCExl3Og8ifwB6"
             crossorigin="anonymous"
     ></script>
+    <script type="text/javascript" src='https://kit.fontawesome.com/a076d05399.js'></script>
```
bkeroackdsc, line 26: Unnecessary javascript if you use a unicode literal (see above).

pkg/api/api_v2.go
esterlinej: Ah, good point!

ui/views/env.html
esterlinej: Thanks, I found it on the font awesome [cheat sheet](https://fontawesome.com/v4.7.0/cheatsheet/)

esterlinej:
@bkeroackdsc Looking for guidance on the modal confirmation messaging
sync-modal
rebuild-modal

bkeroackdsc:
@esterlinej Per our discussion:

> Synchronize will perform Helm upgrades with the current Pull Request code revisions on all the applications in your environment.

> Rebuild will destroy and recreate your environment from scratch, including image builds (if necessary). You will lose any unsaved data in the environment.

esterlinej:
sync
reload

esterlinej:
sync-2

ui/assets/env.js
```diff
@@ -164,5 +164,39 @@ document.addEventListener("DOMContentLoaded", function(){
         e.preventDefault();
         update();
     });
+    if (renderActions === "true") {
```
bkeroackdsc, line 167: Instead of the boolean flag, I would wrap this in a conditional:

```javascript
if document.getElementById("synchronizeModal") !== null {
  // ...
}
```

ui/views/env.html
```diff
@@ -4,6 +4,7 @@
     <script type="text/javascript">
         const apiBaseURL = "{{ .APIBaseURL }}";
         const envName = "{{ .EnvName }}";
+        const renderActions = "{{ .RenderActions }}";
```
bkeroackdsc, line 7: This boolean becomes unnecessary if the JS checks for the existence of elements instead.

pkg/persistence/datalayer_test.go
```diff
@@ -1114,6 +1115,26 @@ func TestDataLayerUpdateK8sEnvTillerAddr(t *testing.T) {
 	}
 }
 
+func TestFakeDataLayerUpdateK8sEnvConfSignature(t *testing.T) {
```
bkeroackdsc, line 1118: Rename to `TestDataLayerUpdateK8sEnvConfSignature` (it doesn't just test the fake, it tests both implementations).

pkg/api/api_ui.go
```diff
@@ -289,13 +289,15 @@ type BaseTemplateData struct {
 	APIBaseURL      string
 	Branding        uiBranding
 	RenderEventLink bool
+	RenderActions   bool
```
bkeroackdsc, line 292: This field is specifically only for the environment detail page, so it shouldn't be part of `BaseTemplateData`. Instead create a new type that embeds `BaseTemplateData` similar to the home page template data: https://github.com/dollarshaveclub/acyl/blob/eddd0a021faa6ad6e8a63085e4c351b909e2fe33/pkg/api/api_ui.go#L661-L664

esterlinej:
Pushed changes per CR, manual test with `mockui` and run datalayer unit tests.

PR #99, opened 2020-07-30
Feature: Namespace Contents

### LINKED ISSUE #85: Namespace contents list in status page
It would be useful for users to have a listing of environment namespace contents in the status page as an environment is created. Pods, Jobs, etc.

esterlinej:
It would be helpful for users to be able to track their environment creation by viewing the environment namespace contents in real time in the event status UI.
- We want to implement a new UI element that displays a continuously-updated list of pods in the environment namespace along with status. The initial feature should be as simple as possible and should mirror the output of kubectl get pods. In future iterations and improvements we can add additional details like pod logs, etc.
Details
- The feature should mirror the default output of kubectl get pods as much as possible, for familiarity and ease of use. We should use a monospace font (similar to the Debug Logs display) and column-based display similar to kubectl.
- We should use the same columns: Name, Ready, Status, Restarts, Age

### UI
We should use an additional accordion panel below the environment tree, and it should be collapsed by default (similar to the other non-tree panels).
- We should make the new panel visually distinguishable by including the blue/white kubernetes icon in the title.

-----------------------------
Fixes #85


pkg/api/api_v2.go
```diff
@@ -829,3 +833,60 @@ func (api *v2api) userEnvActionsRebuildHandler(w http.ResponseWriter, r *http.Re
...
+		k8senv = &models.KubernetesEnvironment{}
+	}
+	pl, err := api.kr.GetK8sEnvPodList(context.Background(), k8senv.Namespace)
+	if err != nil {
+		api.rlogger(r).Logf("error getting pod list for k8s env: %v", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	w.Header().Add("Content-Type", "application/json")
+	if err := json.NewEncoder(w).Encode(&pl); err != nil {
```
bkeroackdsc, line 888: Do not marshal and emit the metahelm type directly. Every API method should define it's own specific type that (once finalized and deployed) is considered immutable and may not be changed later.

For example: https://github.com/dollarshaveclub/acyl/blob/a78af08715dc5c0b52a554e91c4d677555412f76/pkg/api/api_v2.go#L638-L644

pkg/api/api_v2.go
```diff
@@ -829,3 +833,60 @@ func (api *v2api) userEnvActionsRebuildHandler(w http.ResponseWriter, r *http.Re
...
+	k8senv, err := api.dl.GetK8sEnv(r.Context(), envname)
+	if err != nil {
+		api.rlogger(r).Logf("error getting k8s env from db: %v", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	if k8senv == nil {
+		// if there's no k8senv yet, we'll just use an empty one (this will result in an empty namespace in the UI)
+		k8senv = &models.KubernetesEnvironment{}
+	}
```
bkeroackdsc, line 880: If there's no namespace yet, there's no point in continuing. Just return an empty list to the client.

pkg/api/api_v2_test.go
```diff
@@ -30,7 +30,7 @@ func TestAPIv2SearchByTrackingRef(t *testing.T) {
 	}
 	defer tdl.TearDown()
 	rc := httptest.NewRecorder()
-	apiv2, err := newV2API(dl, nil, nil, config.ServerConfig{APIKeys: []string{"foo"}}, OAuthConfig{}, testlogger)
+	apiv2, err := newV2API(dl, nil, nil, config.ServerConfig{APIKeys: []string{"foo"}}, OAuthConfig{}, testlogger, nil)
```
bkeroackdsc, line 33: Add a test for your new API handler.

pkg/nitro/metahelm/metahelm.go
```diff
@@ -1104,3 +1109,51 @@ func (ci ChartInstaller) removeOrphanedCRBs(ctx context.Context, maxAge time.Dur
 	}
 	return nil
 }
+
+type uiPod struct {
```
bkeroackdsc, line 1113: This type should be exported. It doesn't make much sense to have an exported method return an unexported type.

pkg/nitro/metahelm/metahelm.go
```diff
@@ -1104,3 +1109,51 @@ func (ci ChartInstaller) removeOrphanedCRBs(ctx context.Context, maxAge time.Dur
...
+	Name, Ready, Status string
+	Restarts            int32
+	Age                 time.Duration // TODO: change to string, formatted time
```
bkeroackdsc, line 1116: Keep this as a time.Duration and parse to a string either in the API handler, or in the client-side Javascript (whichever is more convenient).

pkg/nitro/metahelm/metahelm.go
bkeroackdsc: Also this type shouldn't have "UI" in the name since it is not UI-specific (anything could use the method to obtain namespace pod information).

pkg/nitro/metahelm/metahelm.go
```diff
@@ -1104,3 +1109,51 @@ func (ci ChartInstaller) removeOrphanedCRBs(ctx context.Context, maxAge time.Dur
...
+		return &corev1.PodList{}, errors.Wrapf(err, "error unable to retrieve pods for namespace %v", ns)
+	}
+	return pl, nil
+}
+
+// GetK8sEnvPodList returns a kubernetes environment per the name requested
+func (ci ChartInstaller) GetK8sEnvPodList(ctx context.Context, ns string) (pods []uiPod, err error) {
+	pl, err := ci.getPodList(ctx, ns)
+	if len(pl.Items) == 0 {
+		return []uiPod{}, errors.Wrapf(err, "error no pods found for namespace %v", ns)
```
bkeroackdsc, line 1131: No pods shouldn't be an error.

pkg/nitro/metahelm/metahelm.go
```diff
@@ -1104,3 +1109,51 @@ func (ci ChartInstaller) removeOrphanedCRBs(ctx context.Context, maxAge time.Dur
...
+	}
+	for _, p := range pl.Items {
+		age := time.Since(p.CreationTimestamp.Time)
+		var nReady int
+		nContainers := len(p.Spec.Containers)
+		if string(p.Status.Phase) != "Running" {
```
bkeroackdsc, line 1137: Use the named constants provided by the k8s packages and not literal strings.

pkg/nitro/metahelm/metahelm.go
```diff
@@ -1104,3 +1109,51 @@ func (ci ChartInstaller) removeOrphanedCRBs(ctx context.Context, maxAge time.Dur
...
+			if p.Kind == "Pod" {
```
bkeroackdsc, line 1138: Unnecessary conditional since you specifically fetched pod objects.

pkg/nitro/metahelm/metahelm.go
```diff
@@ -1104,3 +1109,51 @@ func (ci ChartInstaller) removeOrphanedCRBs(ctx context.Context, maxAge time.Dur
...
 	return nil
 }
+
+type uiPod struct {
+	Name, Ready, Status string
+	Restarts            int32
+	Age                 time.Duration // TODO: change to string, formatted time
+}
+
+func (ci ChartInstaller) getPodList(ctx context.Context, ns string) (*corev1.PodList, error) {
```
bkeroackdsc, line 1119: This method seems unnecessary since it's used in one place and is just a wrapper around a single k8s client API call.

pkg/nitro/metahelm/metahelm.go
```diff
@@ -1104,3 +1109,51 @@ func (ci ChartInstaller) removeOrphanedCRBs(ctx context.Context, maxAge time.Dur
...
+	pl, err := ci.kc.CoreV1().Pods(ns).List(meta.ListOptions{})
+	if err != nil {
+		return &corev1.PodList{}, errors.Wrapf(err, "error unable to retrieve pods for namespace %v", ns)
+	}
+	return pl, nil
+}
+
+// GetK8sEnvPodList returns a kubernetes environment per the name requested
```
bkeroackdsc, line 1127: Determine whether this method assumes the namespace exists (or not), and document it here.

pkg/api/api_v2.go
```diff
@@ -829,3 +833,60 @@ func (api *v2api) userEnvActionsRebuildHandler(w http.ResponseWriter, r *http.Re
...
+	pl, err := api.kr.GetK8sEnvPodList(context.Background(), k8senv.Namespace)
```
bkeroackdsc, line 881: It's not clear if it's safe to call `GetK8sEnvPodList()` if the namespace doesn't exist. Either check first and don't call this if the namespace doesn't exist, or document that the method will succeed and return an empty list if namespace doesn't exist.

pkg/nitro/metahelm/metahelm.go
bkeroackdsc: It's not clear why it would matter if the pod is running or not for the logic that follows this conditional.

pkg/nitro/metahelm/metahelm.go
```diff
@@ -1104,3 +1109,51 @@ func (ci ChartInstaller) removeOrphanedCRBs(ctx context.Context, maxAge time.Dur
...
+		if string(p.Status.Phase) != "Running" {
+			if p.Kind == "Pod" {
+				for _, c := range p.Status.ContainerStatuses {
+					if c.Ready {
+						nReady += 1
+					}
+				}
+			}
+		} else {
+			if p.Kind == "Pod" {
```
bkeroackdsc, line 1146: It's not clear why you're checking for the pod kind again.

esterlinej: In the image example documented in the feature, it displays a Job Ready: 0/1, and if the type is a Job, the current logic will display Ready: 1/1. This is a simple way of only updating nReady if it's an actual Pod.

bkeroackdsc: We shouldn't be displaying any objects other than Pods. Jobs/Deployments/StatefulSets/etc are higher-level objects that create & manage pods.

esterlinej: When listing pods, the Jobs also may be included, but I have an alternate way to manage the display (0/1 Completed) versus what's currently present in the PR. Will update soon.

esterlinej: I will update and remove the Job portions then.

pkg/api/api_v2.go
esterlinej: I updated the previous check to return an empty env response if no env is found, and shouldn't make this call if does not exist. If it does exist and no pods are found, it returns an error.

pkg/nitro/metahelm/metahelm.go
bkeroackdsc: see my previous CR comment: https://github.com/dollarshaveclub/acyl/pull/99#discussion_r464736405

pkg/api/api_v2.go
bkeroackdsc: See my previous CR comment: https://github.com/dollarshaveclub/acyl/pull/99#discussion_r464736405

No pods in an namespace **is not** an error

acyl.yml
```diff
@@ -10,7 +10,8 @@ application:
   image: quay.io/dollarshaveclub/acyl
   value_overrides:
     - "app.ui.enforce_oauth=true"
-    - "ingress.traefik.enabled=false"
+    - "ingress.traefik.enabled=true"
+    - "ingress.traefik.tls_secret_name=cert-dqa-email"
```
esterlinej, line 14: To be removed.

pkg/api/api_v2.go
```diff
@@ -829,3 +833,89 @@ func (api *v2api) userEnvActionsRebuildHandler(w http.ResponseWriter, r *http.Re
...
+	}
+	// Returns error if no pod(s) found for namespace
+	pl, err := api.kr.GetK8sEnvPodList(context.Background(), k8senv.Namespace)
+	if err != nil {
+		api.rlogger(r).Logf("error getting pod list for k8s env: %v", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	for _, p := range pl {
+		age, err := time.ParseDuration(p.Age.String())
```
bkeroackdsc, line 909: `p.Age` is already a time.Duration, this Parse() call is unnecessary.

pkg/api/api_v2.go
```diff
@@ -829,3 +833,89 @@ func (api *v2api) userEnvActionsRebuildHandler(w http.ResponseWriter, r *http.Re
...
+			api.rlogger(r).Logf("error parsing pod age: %v", err)
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+		v2enp = append(v2enp, V2EnvNamePods{
+			Name: p.Name,
+			Ready: p.Ready,
+			Status: p.Status,
+			Restarts: fmt.Sprintf("%v", p.Restarts),
+			Age: age.Round(time.Second).String(),
```
bkeroackdsc, line 920: This should be `p.Age.Round(time.Second).String()`

pkg/api/api_v2.go
```diff
@@ -829,3 +833,89 @@ func (api *v2api) userEnvActionsRebuildHandler(w http.ResponseWriter, r *http.Re
...
+		if err != nil {
+			api.rlogger(r).Logf("error parsing pod age: %v", err)
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+		v2enp = append(v2enp, V2EnvNamePods{
+			Name: p.Name,
+			Ready: p.Ready,
+			Status: p.Status,
+			Restarts: fmt.Sprintf("%v", p.Restarts),
```
bkeroackdsc, line 919: This is fine, though `strconv.Itoa()` would probably be the better choice.

ui/views/status.html
```diff
@@ -64,6 +64,43 @@ <h2 class="mb-0">
...
+                  <button
+                          class="btn btn-link text-dark btn-lg"
+                          type="button"
+                          data-toggle="collapse"
+                          data-target="#collapseK8sNamespacePods"
+                          aria-expanded="true"
+                          aria-controls="collapseK8sNamespacePods"
+                          id="k8sNamespaceButton"
+                  >
+                    Kubernetes Namespace
```
bkeroackdsc, line 79: Put the attached k8s icon (decompress first) in `ui/assets/` and change this line to:

```html
Kubernetes Namespace <img class="mb-1" src="{{ .APIBaseURL }}/ui/static/k8sicon.svg" width="20" height="20">
```

To match the visual style in the mockup.

[k8sicon.svg.gz](https://github.com/dollarshaveclub/acyl/files/5031954/k8sicon.svg.gz)

esterlinej: Thank you for reminding me about this, I had searched for this, but this fell to the side.

pkg/api/api_v2.go
```diff
@@ -829,3 +833,90 @@ func (api *v2api) userEnvActionsRebuildHandler(w http.ResponseWriter, r *http.Re
...
+	}
+	// Return empty kubernetes environment if not found
+	if k8senv == nil {
+		w.Header().Add("Content-Type", "application/json")
+		if err := json.NewEncoder(w).Encode(&v2enp); err != nil {
+			api.rlogger(r).Logf("error marshaling user env detail: %v", err)
+			w.WriteHeader(http.StatusInternalServerError)
+		}
+	}
+	pl, err := api.kr.GetK8sEnvPodList(context.Background(), k8senv.Namespace)
```
bkeroackdsc, line 894: This context should be `r.Context()`.

ui/assets/status.js
```diff
@@ -12,6 +12,7 @@ const pollingIntervalMilliseconds = 750;
 let done = false;
 let failures = 0;
 let updateInterval = setInterval(update, pollingIntervalMilliseconds);
+let env_name = null;
```
bkeroackdsc, line 15: An empty string would be a safer default value.

bkeroackdsc: See below, I don't think you need this global variable.

ui/assets/status.js
```diff
@@ -623,6 +675,29 @@ async function update() {
...
+    if (env_name !== null) {
+        req3.open('GET', `${apiBaseURL}/v2/userenvs/${env_name}/namespace/pods`, true);
+        req3.onload = function (e) {
+            if (req3.status !== 200) {
+                failures++;
+                console.log(`namespace pods request failed (${failures}): ${req3.status}: ${req3.responseText}`);
+                if (failures >= 10) {
+                    console.log(`API failures exceed limit: ${failures}: aborting update`);
+                    clearInterval(updateInterval);
+                }
```
bkeroackdsc, line 688: I don't think we need to track failures for the pod list API call, you could just `console.log()` the status code and response if this call fails.

pkg/api/api_v2.go
```diff
@@ -829,3 +833,90 @@ func (api *v2api) userEnvActionsRebuildHandler(w http.ResponseWriter, r *http.Re
...
+		return
+	}
+	repos, err := userPermissionsClient(api.oauth).GetUserVisibleRepos(r.Context(), uis)
+	if err != nil {
+		api.rlogger(r).Logf("error getting user visible repos: %v: %v", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	if !repoInRepos(repos, qae.Repo) {
+		w.WriteHeader(http.StatusForbidden)
```
esterlinej, line 877: Thoughts on returning an empty pods struct `[]V2EnvNamePods{}` rather than forbidden so at least the headings render on the status page rather than a blank `Kubernetes Namespace` box?

PR #104, opened 2020-07-31
DO NOT MERGE Secret fetcher refactor - first draft

cleanshavenalex:
Adding FileReadSecretsFetcher to support Vault Agent Injector usage moved secrets client logic from root.go to secrets pkg left off at K8sConfig Process Secrets Injections which currently accepts PVC Client


.helm/charts/acyl/templates/cronjob.yaml
```diff
@@ -14,6 +14,15 @@ spec:
             heritage: {{.Release.Service | quote }}
             release: {{.Release.Name | quote }}
             chart: "{{.Chart.Name}}-{{.Chart.Version}}"
+          annotations:
+{{- if eq .Values.vault.agent_injector.enabled true }}
```
mikeykhalil, line 18: I could be wrong, but we may want to consider placing the `annotations` key itself within this `if` block. I haven't tested this, but there might be issues decoding this yaml file if the `annotations` key doesn't contain a value.

pkg/secrets/fileread_secrets_fetcher.go
```diff
@@ -0,0 +1,192 @@
...
+	"github.com/dollarshaveclub/acyl/pkg/config"
+
+	"github.com/pkg/errors"
+)
+
+// ReadFileSecretsFetcher will fetch secrets from filepaths using VaultConfig.SecretsRootPath as the base path
+type ReadFileSecretsFetcher struct {
+	vc *config.VaultConfig
+}
+
```
mikeykhalil, line 19:
```suggestion
var _ SecretsFetcher = &ReadFileSecretsFetcher{} // Ensure this implements the SecretsFetcher interface at build time

```

.helm/charts/acyl/templates/cronjob.yaml
cleanshavenalex: thanks I did add `vault.agent_injector.enabled: false` in the values.yaml
will double check

pkg/secrets/fileread_secrets_fetcher.go
cleanshavenalex: thank you!

pkg/secrets/secrets.go
```diff
@@ -219,3 +223,100 @@ func (psf *PVCSecretsFetcher) PopulateServer(srv *config.ServerConfig) error {
...
+	}
+	err = sf.PopulateAllSecrets(allConfigs.AWSCreds, allConfigs.GithubConfig, allConfigs.SlackConfig, allConfigs.ServerConfig, allConfigs.PGConfig)
+	if err != nil {
+		return errors.Wrap(err, "secrets.PopulateAllSecrets error populating a configuration")
+	}
+	return nil
+}
+
+func newSecretFetcher(secretsBackend string, secretsConfig *config.SecretsConfig, vaultConfig *config.VaultConfig) (SecretFetcher, error) {
+	if vaultConfig.UseAgent {
```
mikeykhalil, line 269: This seems a bit confusing, as I could technically set the `secretsBackend` to "env", and then end up using a `ReadFileSecretsFetcher`, if the vaultConfig.UseAgent is set to `true`.

My initial thought is that perhaps we want a "file" secrets backend? In which case, it might also make sense to remove the `ReadFileSecretsFetcher` dependency on the vault config altogether, and just have its own config struct that just contains the `secretsRootPath`?

cmd/root.go
```diff
@@ -34,6 +31,8 @@ var RootCmd = &cobra.Command{
 //shorthands here: abcdefg n jkl p q
 func init() {
 	RootCmd.PersistentFlags().StringVarP(&vaultConfig.Addr, "vault-addr", "a", os.Getenv("VAULT_ADDR"), "Vault URL (if using Vault secret backend)")
+	RootCmd.PersistentFlags().BoolVar(&vaultConfig.UseAgent, "vault-agent", false, "Use Vault Agent Injector")
```
bkeroackdsc, line 34: Don't create a new flag for the vault agent specifically, since from the application's perspective it's just reading secrets from local files.

cmd/root.go
```diff
@@ -34,6 +31,8 @@ var RootCmd = &cobra.Command{
...
+	RootCmd.PersistentFlags().StringVar(&vaultConfig.SecretsRootPath, "vault-secrets-root-path", "/vault/secrets", "Path where Vault Agent Injector is expected to write secrets ( if vault.UseAgent true )")
```
bkeroackdsc, line 35: `--secrets-root-path`
"Local filesystem root path to read read secrets from"

pkg/secrets/secrets.go
```diff
@@ -219,3 +223,100 @@ func (psf *PVCSecretsFetcher) PopulateServer(srv *config.ServerConfig) error {
 	}
 	return nil
 }
+
+//// Below added to support different fetcher types and consolidate creating the correct fetcher to the secrets package
```
bkeroackdsc, line 227: I understand the motivation for this, but it is unnecessary and duplicates what PVC does internally.

Instead have this logic live in `cmd`, and instantiate either a `PVCSecretsFetcher` or a `LocalFileSecretsFetcher` depending upon the CLI flags passed by the user.

pkg/secrets/fileread_secrets_fetcher.go
```diff
@@ -0,0 +1,192 @@
...
+		return errors.Wrap(err, "error getting server secrets")
+	}
+	if err := rf.PopulatePG(pg); err != nil {
+		return errors.Wrap(err, "error getting db secrets")
+	}
+	return nil
+}
+
+// PopulatePG populates postgres secrets into pg
+func (rf *ReadFileSecretsFetcher) PopulatePG(pg *config.PGConfig) error {
```
bkeroackdsc, line 48: Every one of these methods should check for a nil parameter first (in this case, `pg`).

pkg/secrets/fileread_secrets_fetcher.go
```diff
@@ -0,0 +1,192 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/dollarshaveclub/acyl/pkg/config"
+
+	"github.com/pkg/errors"
+)
+
+// ReadFileSecretsFetcher will fetch secrets from filepaths using VaultConfig.SecretsRootPath as the base path
+type ReadFileSecretsFetcher struct {
```
bkeroackdsc, line 16: This name is awkward. Something like `LocalFileSecretsFetcher` might be better.

pkg/secrets/fileread_secrets_fetcher.go
```diff
@@ -0,0 +1,192 @@
+package secrets
```
bkeroackdsc, line 1: The name of this file should be simplified. Something like `localfile.go`.

pkg/secrets/fileread_secrets_fetcher.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+	if err := rf.PopulatePG(pg); err != nil {
+		return errors.Wrap(err, "error getting db secrets")
+	}
+	return nil
+}
+
+// PopulatePG populates postgres secrets into pg
+func (rf *ReadFileSecretsFetcher) PopulatePG(pg *config.PGConfig) error {
+	s, err := ioutil.ReadFile(fmt.Sprintf("%v/dburid", rf.vc.SecretsRootPath))
```
bkeroackdsc, line 49: File paths shouldn't be hardcoded here. I'm not clear on where the naming scheme is coming from (the chart?).

Note the list of secret ID constants here: https://github.com/dollarshaveclub/acyl/blob/a78af08715dc5c0b52a554e91c4d677555412f76/pkg/secrets/secrets.go#L16-L35

You should utilize them to generate these filesystem paths, instead of hardcoding arbitrary filenames.

pkg/secrets/fileread_secrets_fetcher.go
```diff
@@ -0,0 +1,192 @@
...
+	if err != nil {
+		return errors.Wrap(err, "error getting DB URI")
+	}
+	pg.PostgresURI = string(s)
+	return nil
+}
+
+// PopulateAWS populates AWS secrets into aws
+func (rf *ReadFileSecretsFetcher) PopulateAWS(aws *config.AWSCreds) error {
+	s, err := ioutil.ReadFile(fmt.Sprintf("%v/awsaccesskeyid", rf.vc.SecretsRootPath))
```
bkeroackdsc, line 59: It's probably also worth implementing a [file size sanity check](https://golang.org/pkg/os/#File.Stat), since you're reading these files fully into memory.

ex:
```go
var MaxSecretFileSizeBytes = 1_000_000 // 1 MB```

**Comment by bkeroackdsc** [comment]
@cleanshavenalex Note also that the image build is failing, so that will need to be fixed.

File: .helm/charts/acyl/values.yaml [hunk 12]
```diff
@@ -60,6 +60,49 @@ vault:
   use_k8s_auth: true
   role: acyl
   auth_path: "kubernetes"
+  agent_injector:
+    enabled: false
+    init_first: false
+    log_level: info
+  secrets_prefix: "secret/services/acyl/qa/"
+  secrets:
+  - name: awsAccessKeyID
```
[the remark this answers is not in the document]
cleanshavenalex: I did try to use the path as the secret name and filepath , the way the agent annotations for secret and secret template works is you have to have a secret name and template name that does not have '/' character and believe the secret path 
The name becomes the filename from the `/vault/secrets` root. 
We could potentially use a Helm template function to swap '/' for '-' in the path saved to a new variable and use that. Ill give that a go in the starter templates.

PR #106, opened 2020-08-11
Feature/pod logs

esterlinej:
It would be helpful for users to be able to view pod logs in the status UI, especially for pods with failures (CrashLoopBackoff).

### Details
We should provide a way for users to click the pod name in the Kubernetes Namespace accordion panel and see a display of the last 100 lines of pod log output, along with a way to refresh the log contents.
- In future iterations of this feature we may want to provide an option for the user to specify the max number of log lines, but for this initial iteration it should be set to 100.

### UI
This should be implemented in the following way:
- The pod name in the Kubernetes Namespace panel should be a clickable link.
- Alternatively, this could be an additional column in the pod info table (all the way to the right) with a clickable “Logs” button.
- Whichever is aesthetically cleaner is better. The pod name link option keeps visual parity with kubectl output so is preferable.
- The pod name link should open up a modal, with an internal, scrollable text area with the pod logs. 
- It should be autoscrolled to the bottom (the most recent pod log lines).
- The log display must be a monospace font (the same font as the Debug Logs).
- The log display should be white text on a dark background (it should match the pod list output, visually).
- There should be a “Refresh” button at the bottom of the modal, which should update the log output with the most recent N lines.
- A real time log tail is more complex than we want to tackle for the first iteration of this feature.
- There should be a “Close” button at the bottom of the modal that dismisses it.


pkg/api/api_v2.go
```diff
@@ -902,23 +905,205 @@ func (api *v2api) userEnvNamePodsHandler(w http.ResponseWriter, r *http.Request)
...
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	// Return empty kubernetes environment if not found
+	if k8senv == nil {
+		w.Header().Add("Content-Type", "text/plain")
+		if err := json.NewEncoder(w).Encode(""); err != nil {
+			api.rlogger(r).Logf("error marshaling user env detail: %v", err)
+			w.WriteHeader(http.StatusInternalServerError)
+		}
```
mikeykhalil, line 1068:
```suggestion
```

What does ` json.NewEncoder(w).Encode("")` in this case? Would returning suffice in this case?

pkg/api/api_v2.go
```diff
@@ -902,23 +905,205 @@ func (api *v2api) userEnvNamePodsHandler(w http.ResponseWriter, r *http.Request)
...
+	defer plRC.Close()
+	podLogs, err := io.Copy(w, plRC)
+	if err != nil {
+		api.rlogger(r).Logf("error copying pod logs output")
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	w.Header().Add("Content-Type", "text/plain")
+	if err := json.NewEncoder(w).Encode(&podLogs); err != nil {
+		api.rlogger(r).Logf("error marshaling user env pod logs: %v", err)
```
mikeykhalil, line 1106: Few things:
1. I believe we are writing to the ResponseWriter here: `podLogs, err := io.Copy(w, plRC)`
2. When doing io.Copy(), technically, the ResponseWriter will end up setting the Content-Type header for you, since you don't set it until after the io.Copy call. The ResposneWriter will also call `w.WriteHeader` as well for you as well in the first call to its Write function . Reference:: https://golang.org/pkg/net/http/#ResponseWriter
3. How come we are using a json Encoder to write the value `podLogs`? `podLogs` should be an `int64`, so we should be able to do something like `w.Write([]byte(strconv.FormatInt(podLogs, 10)))`

If I understand this function correctly, my suggestion would be the following
```suggestion
// Set the Content-Type before so the ResponseWriter.Write call doesn't set it for us
w.WriteHeader("Content-Type", "text/plain")

// Copy to a buffer to ensure we know the length of the response
var buf []byte
written, err := io.CopyBuffer(w, plRC, buf)
if err != nil { 
     api.rlogger(r).Logf("error copying pod logs to buffer: %v", err)
     w.WriteHeader(http.StatusInternalServerError)
     return 
}
// Write the content-length in case we run into issues writing all of the logs to the ResponseWriter
w.WriteHeader("Content-Length", written)
_, err := w.Write(buf)
if err != nil { 
    // log error 
    return 
}

```

pkg/api/api_v2.go
```diff
@@ -902,23 +905,205 @@ func (api *v2api) userEnvNamePodsHandler(w http.ResponseWriter, r *http.Request)
...
+	}
+	pl, err := api.kr.GetK8sEnvPodList(r.Context(), k8senv.Namespace)
+	if err != nil {
+		api.rlogger(r).Logf("error getting pod list for k8s env: %v", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	// Return empty pod list if nil
+	if pl == nil {
+		w.Header().Add("Content-Type", "application/json")
```
mikeykhalil, line 992: I would suggest using w.Header().Set() in cases where you want to set a header. I believe Add will append to any existing values, which might cause some unexpected behaviors with the `Content-Type` header.

pkg/nitro/metahelm/metahelm.go
```diff
@@ -1146,3 +1151,74 @@ func (ci ChartInstaller) GetK8sEnvPodList(ctx context.Context, ns string) (out [
...
+	}
+	logRC, err := req.Stream()
+	if err != nil {
+		return rc, errors.Wrap(err, "error getting logs")
+	}
+	return logRC, nil
+}
+
+func lineCounter(body []byte) (uint, error) {
+	buf := make([]byte, 68*1024)
```
mikeykhalil, line 1211: For this particular function, if we already have the whole body contained in a single byte slice, then we should just be able to do 
```
bytes.Count(body, []byte{'\n'})
```

pkg/nitro/metahelm/metahelm.go
```diff
@@ -1146,3 +1151,74 @@ func (ci ChartInstaller) GetK8sEnvPodList(ctx context.Context, ns string) (out [
...
+	req := ci.kc.CoreV1().Pods(ns).GetLogs(podname, &plo)
+	if req == nil {
+		return rc,nil
+	}
+	req.BackOff(nil)
+	body, err := req.Body([]byte{}).DoRaw()
+	if err != nil {
+		return rc, errors.Wrap(err, "error getting body")
+	}
+	lineCount, err := lineCounter(body)
```
mikeykhalil, line 1196: How come we have to verify the number of lines here? Do we not trust the Kubernetes client `PodLogOptions.TailLines`?

Also, given that we are returning the stream, I assume it's possible that the number of lines may continue to grow.

See the kubectl implementation for some more details of how they do it: 

https://github.com/kubernetes/kubernetes/blob/c2e90cd1549dff87db7941544ce15f4c8ad0ba4c/pkg/kubectl/cmd/log.go#L188

esterlinej: Within the feature doc
```
The KubernetesReporter method should use the k8s client GetLogs method. 
The method should do a sanity check on the number of lines requested and 
return an error if it exceeds a reasonable threshold (1000 seems OK to me).
```

mikeykhalil: > sanity check on the number of lines request

Right. But in this example, you are performing a validation on the number of lines _received_ from the kubernetes API. Instead, I would suggest validating this input from the user. Specifically, the value obtained from `r.URL.Query().Get("lines")`

pkg/api/api_v2.go
esterlinej: Meant to come back to this, thank you.

pkg/nitro/metahelm/metahelm.go
esterlinej: HA! I converted this from accepting a Reader to just the byte slice itself, and didn't think of this! Thank you.

pkg/nitro/metahelm/metahelm.go
esterlinej: The good thing now is since we're working off the body/byte slice, we can simply check this without much fuss:
```
lineCount := bytes.Count(buf, []byte{'\n'})
if lineCount > nLogLines {
	// produce error
}
```

pkg/api/api_v2.go
```diff
@@ -902,23 +905,191 @@ func (api *v2api) userEnvNamePodsHandler(w http.ResponseWriter, r *http.Request)
...
+	defer pl.Close()
+	w.Header().Set("Content-Type", "text/plain")
+	var buf []byte
+	written, err := io.CopyBuffer(w, pl, buf)
+	if err != nil {
+		api.rlogger(r).Logf("error copying pod logs to buffer: %v", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	w.Header().Set("Content-Length", strconv.FormatInt(written, 10))
```
bkeroackdsc, line 1089: I understand @mikeykhalil's point about the Content-Length header, but since we don't use or care about it on the client side, I don't see the value in this considering we need to copy the entire logs output to a buffer.

It would be simpler and more elegant to copy the ReadCloser to the ResponseWriter directly.

pkg/nitro/metahelm/metahelm.go
```diff
@@ -1146,3 +1150,50 @@ func (ci ChartInstaller) GetK8sEnvPodList(ctx context.Context, ns string) (out [
...
+// GetK8sEnvPodContainers returns all container names for the specified pod
+func (ci ChartInstaller) GetK8sEnvPodContainers(ctx context.Context, ns, podname string) (out K8sPodContainers, err error) {
+	pod, err := ci.kc.CoreV1().Pods(ns).Get(podname, meta.GetOptions{})
+	if err != nil {
+		return K8sPodContainers{}, errors.Wrapf(err, "error unable to retrieve pods for namespace %v", ns)
+	}
+	if pod == nil {
+		return K8sPodContainers{}, errors.Errorf("error pod %v not found for namespace %v", podname, ns)
+	}
+	containers := []string{}
```
bkeroackdsc, line 1168:
```go
containers := make([]string, len(pod.Spec.Containers))
```

pkg/nitro/metahelm/metahelm.go
```diff
@@ -1146,3 +1150,50 @@ func (ci ChartInstaller) GetK8sEnvPodList(ctx context.Context, ns string) (out [
...
+	return K8sPodContainers{
+		Pod:        pod.Name,
+		Containers: containers,
+	}, nil
+}
+
+// GetK8sEnvPodLogs returns
+func (ci ChartInstaller) GetK8sEnvPodLogs(ctx context.Context, ns, podname, container string, lines uint) (out io.ReadCloser, err error) {
+	rc := ioutil.NopCloser(strings.NewReader(""))
+	if lines > 1000 {
```
bkeroackdsc, line 1181: Define this as a constant: `MaxPodContainerLogLines`

pkg/nitro/metahelm/metahelm.go
```diff
@@ -1146,3 +1150,50 @@ func (ci ChartInstaller) GetK8sEnvPodList(ctx context.Context, ns string) (out [
...
+	}
+	return K8sPodContainers{
+		Pod:        pod.Name,
+		Containers: containers,
+	}, nil
+}
+
+// GetK8sEnvPodLogs returns
+func (ci ChartInstaller) GetK8sEnvPodLogs(ctx context.Context, ns, podname, container string, lines uint) (out io.ReadCloser, err error) {
+	rc := ioutil.NopCloser(strings.NewReader(""))
```
bkeroackdsc, line 1180: Not needed, delete.

pkg/nitro/metahelm/metahelm.go
```diff
@@ -1146,3 +1150,50 @@ func (ci ChartInstaller) GetK8sEnvPodList(ctx context.Context, ns string) (out [
...
+	if lines > 1000 {
+		return rc, errors.Errorf("error line request exceeds limit")
```
bkeroackdsc, line 1182: Return `nil` for the ReadCloser if an error happens. All interface values can be nil.

ui/views/status.html
```diff
@@ -96,6 +96,79 @@ <h2 class="mb-0">
...
+                                            <div class="dropdown">
+                                              <button
+                                                      class="btn btn-secondary dropdown-toggle"
+                                                      type="button"
+                                                      id="containersDropdownMenu"
+                                                      data-toggle="dropdown"
+                                                      aria-haspopup="true"
+                                                      aria-expanded="false"
+                                              >
+                                                Containers Dropdown
```
bkeroackdsc, line 127: Change to "Containers" or "Pod Containers".

pkg/api/api_v2.go
```diff
@@ -902,23 +905,188 @@ func (api *v2api) userEnvNamePodsHandler(w http.ResponseWriter, r *http.Request)
...
+// userEnvNamePodContainersHandler returns list of container names for specified pod and environment name
+func (api *v2api) userEnvNamePodContainersHandler(w http.ResponseWriter, r *http.Request) {
+	v2enpc := V2EnvNamePodContainers{}
+	uis, err := getSessionFromContext(r.Context())
+	if err != nil {
+		api.rlogger(r).Logf("session missing from context")
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+	podname := mux.Vars(r)["pod_name"]
```
bkeroackdsc, line 943: Be consistent with the mux var names. Either use "pod", or change the env name var to "env_name".

pkg/api/api_v2.go
```diff
@@ -301,6 +302,8 @@ func (api *v2api) register(r *muxtrace.Router) error {
 	r.HandleFunc("/v2/userenvs/{name}", middlewareChain(api.userEnvDetailHandler, sessionAuthMiddleware.sessionAuth)).Methods("GET")
 	r.HandleFunc("/v2/userenvs/{name}/actions/rebuild", middlewareChain(api.userEnvActionsRebuildHandler, sessionAuthMiddleware.sessionAuth)).Methods("POST")
 	r.HandleFunc("/v2/userenvs/{name}/namespace/pods", middlewareChain(api.userEnvNamePodsHandler, sessionAuthMiddleware.sessionAuth)).Methods("GET")
+	r.HandleFunc("/v2/userenvs/{name}/namespace/pod/{pod_name}/containers", middlewareChain(api.userEnvNamePodContainersHandler, sessionAuthMiddleware.sessionAuth)).Methods("GET")
```
bkeroackdsc, line 305: I would name the handlers `userEnvPodContainersHandler` and `userEnvPodLogsHandler`.

pkg/api/api_v2.go
```diff
@@ -902,23 +905,188 @@ func (api *v2api) userEnvNamePodsHandler(w http.ResponseWriter, r *http.Request)
...
+	repos, err := userPermissionsClient(api.oauth).GetUserVisibleRepos(r.Context(), uis)
+	if err != nil {
+		api.rlogger(r).Logf("error getting user visible repos: %v: %v", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	if !repoInRepos(repos, qae.Repo) {
+		w.WriteHeader(http.StatusForbidden)
+		return
+	}
```
bkeroackdsc, line 973: Do this check before checking if the environment exists. Always check request permissions first before anything else.

pkg/api/api_v2.go
```diff
@@ -902,23 +905,188 @@ func (api *v2api) userEnvNamePodsHandler(w http.ResponseWriter, r *http.Request)
...
+			api.rlogger(r).Logf("error marshaling user env pod list: %v", err)
+			w.WriteHeader(http.StatusInternalServerError)
+		}
+		return
+	}
+	v2enpc = V2EnvNamePodContainers{
+		Name: podname,
+		Containers: podContainers.Containers,
+	}
+	w.Header().Set("Content-Type", "application/json")
```
bkeroackdsc, line 1002: This is repeated in multiple places. Set the Content-Type header once towards the beginning of the handler.

pkg/api/api_v2.go
```diff
@@ -902,23 +905,188 @@ func (api *v2api) userEnvNamePodsHandler(w http.ResponseWriter, r *http.Request)
...
+		return
+	}
+	podContainers, err := api.kr.GetK8sEnvPodContainers(r.Context(), k8senv.Namespace, podname)
+	if err != nil {
+		w.Header().Set("Content-Type", "application/json")
+		if err := json.NewEncoder(w).Encode(&v2enpc); err != nil {
+			api.rlogger(r).Logf("error marshaling user env pod list: %v", err)
+			w.WriteHeader(http.StatusInternalServerError)
+		}
+		return
```
bkeroackdsc, line 996: I'm guessing you're doing this to handle no pods gracefully, but it would be better to not return an error from `GetK8sEnvPodContainers()` if there are no pods (just an empty pod list), then this conditional could just return a 500 error.

pkg/api/api_v2.go
```diff
@@ -902,23 +905,188 @@ func (api *v2api) userEnvNamePodsHandler(w http.ResponseWriter, r *http.Request)
...
+	repos, err := userPermissionsClient(api.oauth).GetUserVisibleRepos(r.Context(), uis)
+	if err != nil {
+		api.rlogger(r).Logf("error getting user visible repos: %v: %v", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	if !repoInRepos(repos, qae.Repo) {
+		w.WriteHeader(http.StatusForbidden)
+		return
+	}
```
bkeroackdsc, line 1047: Check request permissions first before just about any other handler logic.

pkg/api/api_v2.go
```diff
@@ -902,23 +905,188 @@ func (api *v2api) userEnvNamePodsHandler(w http.ResponseWriter, r *http.Request)
...
+		api.rlogger(r).Logf("error getting k8s env from db: %v", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	if k8senv == nil {
+		api.rlogger(r).Logf("error getting k8s env from db: %v", err)
+		w.WriteHeader(http.StatusPreconditionFailed)
+		return
+	}
+	lines := metahelm.MinPodContainerLogLines
```
bkeroackdsc, line 1059: This should be called `DefaultPodContainerLogLines` and should be a constant in this API package, not the metahelm package.

pkg/api/api_v2.go
```diff
@@ -902,23 +905,188 @@ func (api *v2api) userEnvNamePodsHandler(w http.ResponseWriter, r *http.Request)
...
+		// return error if request exceeds limit
+		if i > metahelm.MaxPodContainerLogLines {
+			api.rlogger(r).Logf("error requested lines exceed limit: %v", i)
+			w.WriteHeader(http.StatusBadRequest)
+			return
+		}
+		// override the minimum lines if request is greater, not allowing negative/invalid requests through
+		if i > lines {
+			lines = i
+		}
```
bkeroackdsc, line 1076: Keep it simple. Just check for `lines < 0`.

pkg/api/api_v2.go
```diff
@@ -902,23 +905,188 @@ func (api *v2api) userEnvNamePodsHandler(w http.ResponseWriter, r *http.Request)
...
+	}
+	container := r.URL.Query().Get("container")
```
bkeroackdsc, line 1078: If the container query param is required, check for an empty string here.

pkg/api/api_v2.go
```diff
@@ -902,23 +905,188 @@ func (api *v2api) userEnvNamePodsHandler(w http.ResponseWriter, r *http.Request)
...
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	defer pl.Close()
+	_, err = io.Copy(w, pl)
+	if err != nil {
+		api.rlogger(r).Logf("error copying pod logs output")
 		w.WriteHeader(http.StatusInternalServerError)
 	}
+	w.Header().Set("Content-Type", "text/plain")
```
bkeroackdsc, line 1091: Set this before the `io.Copy()` call.

pkg/api/api_v2.go
bkeroackdsc: You can combine with the conditional above:

```go
if i < 0 || i > metahelm.MaxPodContainerLogLines {
   // ...
}
```

pkg/nitro/metahelm/fake.go
```diff
@@ -160,4 +181,78 @@ func stubPodData(ns string) *v1.PodList {
...
+	scanner := bufio.NewScanner(br)
+	scanner.Split(bufio.ScanLines)
+	for scanner.Scan() {
+		bucket = append(bucket, scanner.Text())
+	}
+	if err := scanner.Err(); err != nil {
+		return []byte{}, errors.Wrapf(err, "error scanning log lines")
+	}
+	// create random value to vary logs returned for testing
+	r := len(bucket) - rand.Intn(50)
```
bkeroackdsc, line 240: It probably doesn't matter here, but be aware that since you're using `math/rand` and you didn't seed with any random value, this will always return a deterministic sequence of "random" integers.

pkg/nitro/metahelm/metahelm.go
```diff
@@ -64,6 +65,8 @@ type Installer interface {
 // KubernetesReporter describes an object that returns k8s environment data
 type KubernetesReporter interface {
 	GetK8sEnvPodList(ctx context.Context, ns string) (out []K8sPod, err error)
+	GetK8sEnvPodContainers(ctx context.Context, ns, podname string) (out K8sPodContainers, err error)
+	GetK8sEnvPodLogs(ctx context.Context, ns, podname, container string, lines uint) (out io.ReadCloser, err error)
```
bkeroackdsc, line 69: We should drop the "K8sEnv" from all these method names.

pkg/nitro/metahelm/metahelm.go
```diff
@@ -79,6 +82,8 @@ const (
 	DefaultTillerDeploymentName          = "tiller-deploy"
 	DefaultTillerServerConnectRetryDelay = 10 * time.Second
 	DefaultTillerServerConnectRetries    = 40
+	MinPodContainerLogLines 			 = 100
```
bkeroackdsc, line 85: `DefaultPodContainerLogLines`

pkg/nitro/metahelm/metahelm.go
```diff
@@ -1146,3 +1151,51 @@ func (ci ChartInstaller) GetK8sEnvPodList(ctx context.Context, ns string) (out [
...
+	}
+	tl := int64(lines)
+	plo := corev1.PodLogOptions{
+		Container: container,
+		TailLines: &tl,
+	}
+	req := ci.kc.CoreV1().Pods(ns).GetLogs(podname, &plo)
+	if req == nil {
+		return nil,nil
+	}
```
bkeroackdsc, line 1194: Never return nil without an error. This should be an error like "pod logs request is nil".

pkg/nitro/metahelm/metahelm.go
```diff
@@ -1146,3 +1151,51 @@ func (ci ChartInstaller) GetK8sEnvPodList(ctx context.Context, ns string) (out [
...
+	req.BackOff(nil)
+	plRC, err := req.Stream()
+	if err != nil {
+		return nil, errors.Wrap(err, "error getting logs")
```
bkeroackdsc, line 1198: This would be a better error message: "error getting request stream".

ui/assets/status.js
```diff
@@ -552,20 +556,136 @@ function updateLogs(logs) {
...
+            active_container = "";
+            document.getElementById('podLogModalHeading').innerHTML = `<strong>Logs: </strong>`;
+            renderContainerMenuOptions([]);
+            renderPodLogs("Please select a container and click Refresh.");
+        });
+        $("#podLogModalRefresh").on('click', function (e) {
+            e.preventDefault();
+            getPodLogs();
+        });
+    }
```
bkeroackdsc, line 686: I would probably add this to the existing `DOMContentLoaded` event listener, unless that looks totally unreadable.

ui/views/partials/header.html
```diff
@@ -67,6 +67,9 @@
         .acyl-tree {
             cursor: default;
         }
+        .pod-logs {
+            font-family: monospace;
+        }
```
bkeroackdsc, line 72: This is a global header partial. Put this style in the status-specific CSS here (at the top): https://github.com/dollarshaveclub/acyl/blob/master/ui/assets/status.css

ui/views/status.html
```diff
@@ -96,6 +96,83 @@ <h2 class="mb-0">
                       <tbody id="k8sNamespacePodTableBody">
                       </tbody>
                     </table>
+                    <small>
```
bkeroackdsc, line 99: Why is there an inline style tag here? Seems unneeded.

ui/views/status.html
```diff
@@ -96,6 +96,83 @@ <h2 class="mb-0">
...
+                        >
+                          <div class="modal-dialog modal-lg modal-dialog-scrollable" role="document">
+                            <div class="modal-content text-light bg-dark">
+                              <div class="modal-header">
+                                <div class="container">
+                                  <div class="row">
+                                    <div class="col-sm">
+                                      <label id="podLogModalLabel">
+                                        <h6 class="modal-title overflow-auto" id="podLogModalHeading">
+                                          <strong>Logs:</strong>
```
bkeroackdsc, line 117: This is already an `h6` so I don't think you need the explicit strong tag. Regardless, a CSS style would be better.

ui/views/status.html
```diff
@@ -96,6 +96,83 @@ <h2 class="mb-0">
...
+                                >
+                                  <span aria-hidden="true">&times;</span>
+                                </button>
+                              </div>
+                              <div class="modal-body overflow-auto pod-logs">
+                                <div
+                                        class="container-body p4"
+                                        id="podLogsBody"
+                                >
+                                  Please select a container and click Refresh.
```
bkeroackdsc, line 152: I don't think this is visible right? Unneeded anyway, the UI should be self-explanatory.

pkg/api/api_v2.go
esterlinej: I didn't want to change `name` to be consistent with other endpoints, I will use `pod` in place of `pod_name`

pkg/api/api_v2.go
esterlinej: Not sure I understand, `envname` used by `api.dl.GetQAEnvironment(r.Context(), envname)` to then use the response with `!repoInRepos(repos, qae.Repo)`, followed by `GetK8sEnv`. We do this in multiple handler functions.

bkeroackdsc: My point is that handlers should always do permissions checks first, before anything else. Otherwise you're potentially leaking information to an unauthenticated requestor.

ui/views/status.html
esterlinej: I'm having trouble getting the css to work within the status.html, additionally, h6 is a small heading and doesn't include bold, which I wanted for this.

ui/views/status.html
esterlinej: It would only be visible in place of the log not loading, I will remove.

ui/views/partials/header.html
esterlinej: This is the only way I have been able to get this to work. It seems that within the `status.html`, the styling within `status.css` are not reachable.
```
    <link
            href="{{ .APIBaseURL }}/ui/static/status.css"
            rel="stylesheet"
    />
    <style>
      .pod-logs-heading {
        font-weight: bold;
      }

      .pod-logs {
        font-family: monospace;
        font-size: small;
      }
    </style>
```

ui/views/status.html
esterlinej: This was to ensure the font was small, moved to css.

pkg/api/api_v2.go
esterlinej: The container is not required, and though we set the active container, when not specified it defaults to the first, otherwise to the only:
```
// PodLogOptions is the query options for a Pod's logs REST call.
type PodLogOptions struct {
	metav1.TypeMeta `json:",inline"`

	// The container for which to stream logs. Defaults to only container if there is one container in the pod.
	// +optional
	Container string `json:"container,omitempty" protobuf:"bytes,1,opt,name=container"`
```

pkg/api/api_v2.go
esterlinej: Sure, understood. This follows the already established paradigm. Since we need the `qae.Repo` returned from that first check to compare with the list of repositories authorized versus the repo for this environment, I'm not certain how to resolve your request.

ui/views/partials/header.html
bkeroackdsc: OK, not critical. We can come back to this.

PR #108, opened 2020-08-19
Remove consul dependency

mikeykhalil:
Remove consul dependency to make Acyl easier to install and maintain. 

This PR includes a new interface called a `PreemptableLock` and provides a fake implementation and an implementation using postgres advisory locks and listen/notify. 

At first, I made the `PreemptiveLocker` and `PreemptableLocks` accept a single key string as an argument to determine what to lock on. This made the interface a bit more flexible, but didn't play very well with postgres advisory locks, which accept either a single 64 bit integer or 2 32 bit integers as arguments ([reference](https://www.postgresql.org/docs/9.4/functions-admin.html#FUNCTIONS-ADVISORY-LOCKS)). I originally went with a fnv-1a hash to compute the `Sum32` of the input key string, but that still left room for collisions. 

So I decided to just have these structs and interfaces accept 2 32-bit integers as keys, to play nicely with postgres advisory locks and to avoid collisions. This seems to fit well with the main use case of the locker (locking on a repo/pr combination) and can still be extended to serve other purposes (e.g. using 2 unique values for the reaper which requires a lock).


pkg/locker/postgres_lock.go
```diff
@@ -0,0 +1,310 @@
...
+	postgresURI string
+
+	// listener represents a Postgres listener, which watches for Notifications over a defined channel.
+	listener *pq.Listener
+
+	// preempted is a channel which contains payloads from the Postgres Listener.
+	preempted chan NotificationPayload
+
+	// message represents a message to use when notifying other lock holders that their operation has been preempted
+	message string
```
bkeroackdsc, line 113: "message represents a descriptive reason to communicate to other lock holders why their operation was preempted, optional"

pkg/locker/postgres_lock.go
```diff
@@ -0,0 +1,310 @@
...
+				}
+				return
+			}
+			if np.ID == pl.id {
+				pl.log(ctx, "received our own notification, ignoring")
+				continue
+			}
+			// We have received a legimate notification and the lock has been preempted.
+			pl.preempted <- np
+			return
```
bkeroackdsc, line 198:
```go
if np.ID != pl.id {
   pl.preempted <- np
   return
}
```

pkg/locker/locker.go
```diff
@@ -0,0 +1,167 @@
...
+	lock, err := p.lp.AcquireLock(ctx, p.key1, p.key2, event)
+	if err != nil {
+		return nil, fmt.Errorf("unable to acquire lock")
+	}
+
+	p.lock = lock
+	err = p.lock.Notify(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("unable to notify other locks: %v", err)
+	}
```
bkeroackdsc, line 140: `Notify()` should be an internal detail of `lock.Lock()`.

pkg/locker/locker.go
```diff
@@ -0,0 +1,167 @@
...
+
+	ch, err = lock.Lock(ctx, p.opts.LockWait)
+	if err != nil {
+		return nil, fmt.Errorf("unable to lock: %v", err)
+	}
+
+	// Wait for the specified LockDelay before returning
+	time.Sleep(p.opts.LockDelay)
```
bkeroackdsc, line 148: This seems sketchy...

pkg/locker/locker.go
bkeroackdsc: Notification is an implementation detail of the Postgres locker specifically, and thus should not be used by the LockProvider-agnostic PreemptiveLocker.

pkg/locker/locker.go
```diff
@@ -0,0 +1,167 @@
...
+)
+
+var (
+	defaultLockWaitTime = (30 * time.Minute) + (30 * time.Second) // global async timeout is 30 minutes
+	defaultLockDelay    = 15 * time.Second
+)
+
+// LockProvider describes an object capable of creating locks
+type LockProvider interface {
+	AcquireLock(ctx context.Context, key1, key2 int32, event string) (PreemptableLock, error)
```
bkeroackdsc, line 22: I understand why you put `key1, key2 int32` as part of this signature, and it may be unavoidable (unfortunately). But I don't like it because it's allowing an implementation detail of the Postgres lock to leak into this abstract interface definition.

pkg/locker/postgres_lock.go
```diff
@@ -0,0 +1,310 @@
...
+	}
+}
+
+// run is a blocking function that periodically checks the health of the connection
+func (psc *postgresSessionController) run(ctx context.Context) {
+	t := time.NewTicker(psc.keepAlivePeriod)
+	for {
+		select {
+		case <-ctx.Done():
+			psc.sessionErr <- ctx.Err()
```
bkeroackdsc, line 63: Since `psc.sessionErr` is unbuffered, this could block forever and leak the goroutine. I would just return.

In general, if the context is cancelled, abort as quickly as possible.

pkg/locker/postgres_lock.go
```diff
@@ -0,0 +1,310 @@
...
+)
+
+const (
+	defaultKeepAlivePeriod  = 5 * time.Second
+	defaultFailureThreshold = 2
+)
+
+type postgresSessionController struct {
+	// failureThreshold is the number of consecutive failures before marking connection as failed
+	failureThreshold int
```
bkeroackdsc, line 27: unsigned int

pkg/locker/postgres_lock.go
```diff
@@ -0,0 +1,310 @@
...
+
+	// failreCount is the current count of consecutive failures
+	failureCount int
```
bkeroackdsc, line 30: unsigned int

pkg/locker/postgres_lock.go
```diff
@@ -0,0 +1,310 @@
...
+	if keepAlivePeriod == time.Duration(0) {
+		keepAlivePeriod = defaultKeepAlivePeriod
+	}
+
+	if failureThreshold == 0 {
+		failureThreshold = defaultFailureThreshold
+	}
+	return &postgresSessionController{
+		keepAlivePeriod: keepAlivePeriod,
+		conn:            conn,
```
bkeroackdsc, line 52: Check if conn is nil.

pkg/locker/postgres_lock.go
```diff
@@ -0,0 +1,310 @@
...
+	}
+}
+
+func (psc *postgresSessionController) sendKeepAlive(ctx context.Context) error {
+	return psc.conn.PingContext(ctx)
+}
+
+var _ PreemptableLock = &postgresLock{}
+
+type postgresLock struct {
```
bkeroackdsc, line 85: This can probably be exported.

pkg/locker/postgres_lock.go
```diff
@@ -0,0 +1,310 @@
...
+	notificationChannel string
+}
+
+func newPostgresLock(ctx context.Context, db *sqlx.DB, key1, key2 int32, connInfo, message string) (pl *postgresLock, err error) {
+	id, err := uuid.NewUUID()
+	if err != nil {
+		return nil, err
+	}
+
+	conn, err := db.Conn(ctx)
```
bkeroackdsc, line 125: Check db for nil first.

pkg/locker/postgres_lock.go
```diff
@@ -0,0 +1,310 @@
...
+	message string
+
+	// notificationChannel is the listen/notify channel for this lock
+	notificationChannel string
+}
+
+func newPostgresLock(ctx context.Context, db *sqlx.DB, key1, key2 int32, connInfo, message string) (pl *postgresLock, err error) {
+	id, err := uuid.NewUUID()
+	if err != nil {
+		return nil, err
```
bkeroackdsc, line 122: Wrap the error.

pkg/locker/postgres_lock.go
```diff
@@ -0,0 +1,310 @@
...
+	}
+
+	conn, err := db.Conn(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("unable to obtain connection from pool: %v", err)
+	}
+	defer func() {
+		if err != nil {
+			conn.Close()
```
bkeroackdsc, line 131: Check that conn is not nil.

pkg/locker/postgres_lock.go
```diff
@@ -0,0 +1,310 @@
...
+
+func newPostgresLock(ctx context.Context, db *sqlx.DB, key1, key2 int32, connInfo, message string) (pl *postgresLock, err error) {
+	id, err := uuid.NewUUID()
+	if err != nil {
+		return nil, err
+	}
+
+	conn, err := db.Conn(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("unable to obtain connection from pool: %v", err)
```
bkeroackdsc, line 127: `%w` is the error wrap verb.

bkeroackdsc: Though in the rest of the codebase we use pkg/errors Wrap, so that's probably the best thing to use for consistency.

pkg/locker/postgres_lock.go
```diff
@@ -0,0 +1,310 @@
...
+	return pl, nil
+}
+
+// handleEvents is a blocking function.
+// It checks the multiple different channels to determine how to proceed
+func (pl *postgresLock) handleEvents(ctx context.Context, listener *pq.Listener) {
+	for {
+		select {
+		case <-ctx.Done():
+			pl.preempted <- NotificationPayload{
```
bkeroackdsc, line 161: This channel is also unbuffered so this send could block forever.

pkg/locker/postgres_lock.go
```diff
@@ -0,0 +1,310 @@
...
+				pl.log(ctx, "received nil notiifcation")
+				return
+			}
+			payload := notification.Extra
+			np := NotificationPayload{}
+			if err := json.Unmarshal([]byte(payload), &np); err != nil {
+				pl.log(ctx, "could not unmarshal notification payload: %v", err)
+				// In the event that we receive an unknown notification payload, we will give up the lock.
+				// This could help for debugging since we can execute a Notify query to force the app to release the lock.
+				// It could also help if we accidentally make a breaking change to the Notification payload.
```
bkeroackdsc, line 185: 👍 When this is merged, `NotificationPayload` will effectively become an API boundary so we want to be very careful about breaking changes, or relying on the schema for anything.

pkg/locker/locker.go
```diff
@@ -0,0 +1,167 @@
...
+	if p.opts.EnableTracing {
+		return tracer.StartSpanFromContext(ctx, operationName, tracer.ServiceName(p.opts.DatadogServiceName))
+	}
+	// return no-op span if tracing is disabled
+	span, _ := tracer.SpanFromContext(context.Background())
+	return span, ctx
+}
+
+// Lock locks the lock and returns a channel used to signal if the lock should be released ASAP. If the lock is currently in use, this method will block until the lock is released. If caller is preempted while waiting for the lock to be released,
+// an error is returned.
```
bkeroackdsc, line 114: Consider returning a named error for the case where the caller is preempted waiting for the lock. `ErrLockPreempted` or similar.

pkg/locker/lock_test.go
```diff
@@ -0,0 +1,140 @@
...
+	tests := []struct {
+		name  string
+		tfunc func(*testing.T, LockProvider)
+	}{
+		{
+			name:  "lock and unlock",
+			tfunc: testLockAndUnlock,
+		},
+		{
+			name:  "prememption",
```
bkeroackdsc, line 48: typo

pkg/locker/locker.go
mikeykhalil: After doing a bit more digging, I found out using these 2 int32 keys still leaves room for collisions. 

The reason being that Github Repo IDs appear to be [int64 values](https://github.com/google/go-github/issues/597) And since we are using the Repo ID as one of these keys, there will be room for collision on that key. Albeit, it will be fairly unlikely for that collision to occur. 

The options that I see are:
1. Cast int64 Repo ID to int32 and accept the unlikely collision (unlikely collision, slightly messy interface)
2. Just use a single string as a key and utilize a hash function to calculate (unlikely collision, easy to use interface)
3. Insert some int64 ID to uniquely identify a Repo/PR environment within Acyl, and use this as the key (no collisions)

What are your thoughts on which seem to make the most sense for Acyl? I'm thinking option 3, but would understand if we don't want to bother adding new identifiers for Acyl and are not too concerned about the occasional collision.

bkeroackdsc: Yes I agree, 3 makes sense. 

The interface should accept some generic data structure (like a byteslice, or maybe the repo/PR combo) and then the postgres implementation should internally map that to a unique int64 value (storing the mapping in a new table, perhaps) which is then used internally for the advisory lock and notification channel.

pkg/locker/locker.go
```diff
@@ -0,0 +1,167 @@
...
+type NotificationPayload struct {
+	// ID is required so that we can ensure the message came from another party.
+	ID uuid.UUID `json:"id"`
+
+	// Message provides some context as to why the notification was generated.
+	// Useful for logging purposes.
+	Message string `json:"event"`
+
+	//Preempted represents if this notification was instigated because the lock holder has been preempted.
+	Preempted bool `json:"preempted"`
```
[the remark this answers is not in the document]
mikeykhalil: You're right. I was trying to have a field that represented "this operation was preempted by a separate process", just to help for logging and debugging purposes. 

I will think about this a bit more. I have added the lock key to the Notification, and it already contains a UUID, so that might be enough for determining this is from another process. In any case, I will play around with this a bit more.

pkg/locker/postgres_lock.go
```diff
@@ -0,0 +1,310 @@
...
+	// We want to use a single connection as the Advisory Lock we will be using will be a session-level lock
+	// This means that the lock is released in 2 conditions:
+	// 1. The lock is explicitly released (e.g. via pg_advisory_unlock(key bigint))
+	// 2. The session ended (i.e. the tcp connection terminated)
+	// Using a single connection will allow us to detect if the session has ended more reliably than using a connection pool
+	// conn is the single connection to postgres that will be used for the lock
+	conn *sql.Conn
+
+	// postgresURI stores the postgres connection string.
+	postgresURI string
```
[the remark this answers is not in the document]
mikeykhalil: Yes, unfortunately we need the posgresURI as this is a required argument for the listener. PQ establishes a new connection for the listener. Then the other connection is used for locking and unlocking.

pkg/locker/locker.go
mikeykhalil: I think `Notify` might have been an unfortunate name for this method, since it collides with the Postgres command `NOTIFY`. The Notify method for a preemptable lock signals to the process that actually has the corresponding lock know, to release the lock.

Given that it doesn't directly correspond to the PG Notify command, I think it makes sense for the `PreemptableLock` interface to expose this method and for `PreemptiveLocker` to determine when this is called. 

A `PreemptableLock` is essentially a lock that _can_ be preempted , but it can also just be used as a traditional distributed lock. Exposing the `Notify` method makes this possible. For example, in the `reap` package, we do not use a `PreemptiveLocker`, which just directly use the `PreemptableLock` and call the `Lock` method without calling Notify.

pkg/locker/locker.go
mikeykhalil: I agree that a sleep is not ideal and that perhaps we should have some safe guards around the input LockDelay value, but i think that it provides value. 

For context, this is directly inspired by Consul's lock-delay value, which just happens to be implemented on the server side instead of the app side:
```
The final nuance is that sessions may provide a lock-delay. This is a time duration, between 0 and 60 seconds. When a session invalidation takes place, Consul prevents any of the previously held locks from being re-acquired for the lock-delay interval; this is a safeguard inspired by Google's Chubby. The purpose of this delay is to allow the potentially still live leader to detect the invalidation and stop processing requests that may lead to inconsistent state. While not a bulletproof method, it does avoid the need to introduce sleep states into application logic 
```
[reference](https://www.consul.io/docs/internals/sessions)

pkg/locker/postgres_lock.go
mikeykhalil: Ended up changing this to a buffered channel to avoid needing to place some kind of "send this over a channel or timeout after 5 seconds" call throughout this method.

pkg/locker/locker.go
bkeroackdsc: Got it, ok. That makes sense.

pkg/models/models.go
```diff
@@ -374,3 +374,18 @@ type EnvSearchParameters struct {
 	CreatedSince time.Duration       // Duration prior to time.Now().UTC()
 	TrackingRef  string
 }
+
+// EnvLock models the lock
+type EnvLock struct {
```
bkeroackdsc, line 379: Since this is specific to a Postgres lock, I would probably rename to "PostgresEnvLock" or similar.

pkg/nitro/env/env.go
```diff
@@ -215,37 +215,36 @@ func (m *Manager) setGithubCommitStatus(ctx context.Context, rd *models.RepoRevi
...
+	lock := locker.NewPreemptiveLocker(m.LP, repoID, pr, m.PLO)
+	preempt, err := lock.Lock(ctx, "event") // TODO: consider adding more detailed event information
 	if err != nil {
 		end("success:false")
 		return errors.Wrap(err, "error getting lock")
 	}
 	end("success:true")
-	defer lock.Release(ctx)
-
+	defer lock.Release(context.Background())
```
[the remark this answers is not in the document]
bkeroackdsc: Not resolved.

pkg/persistence/aliases.go
```diff
@@ -14,6 +14,7 @@ type RepoRevisionData = models.RepoRevisionData
 type RefMap = models.RefMap
 type QADestroyReason = models.QADestroyReason
 type QAEnvironmentEvent = models.QAEnvironmentEvent
+type EnvLock = models.EnvLock
```
bkeroackdsc, line 17: These type aliases were only used when the code was refactored into individual packages. No new aliases should be created.

pkg/persistence/datalayer.go
```diff
@@ -14,6 +14,7 @@ type LogFunc func(string, ...interface{})
 
 // DataLayer describes an object that interacts with the persistant data store
 type DataLayer interface {
+	CreateEnvLockIfNotExists(ctx context.Context, repo string, pullRequest uint) (el *EnvLock, err error)
```
bkeroackdsc, line 17: If this is only used internally by the Postgres lock implementation (and will only ever be used there), this doesn't need to be in the DataLayer.

The postgres locker has its own Postgres connection so you're not gaining anything by putting this here.

pkg/persistence/pg.go
```diff
@@ -72,6 +72,52 @@ func (p *PGLayer) DB() *sqlx.DB {
...
+		if err != nil {
+			rerr := txn.Rollback()
+			if rerr != nil {
+				p.logger.Printf("error rolling back: %v", rerr)
+			}
+		}
+	}()
+
+	// We don't actually know if the row exists, so we can't rely upon row-level locks to perform this transaction safely
+	_, err = txn.ExecContext(ctx, `LOCK TABLE env_locks`)
```
bkeroackdsc, line 95: This feels too complex. If you have a unique constraint on the lock_key column, you could just attempt the insert and if it works, you're good and if it fails you know there's a collision or some other problem.

pkg/locker/locker_test.go
```diff
@@ -0,0 +1,139 @@
+package locker
```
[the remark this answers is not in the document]
mikeykhalil: Good call. Done.

pkg/nitro/env/env.go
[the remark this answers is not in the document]
mikeykhalil: Whoops, sorry about that! Should be resolved now.

pkg/persistence/datalayer.go
mikeykhalil: Agreed. Moved this to the locker package. Specifically, the LockProvider interface.

pkg/locker/postgres_lock.go
```diff
@@ -0,0 +1,316 @@
...
+	defer func() {
+		// Even if we fail to unlock via Postgres properly, destroying the lock should close the underlying sql.Conn.
+		// At that point, Postgres should clean up the connection.
+		destroyCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		pl.destroy(destroyCtx)
+		cancel()
+	}()
+
+	q := `SELECT pg_advisory_unlock($1)`
+	// TODO (mk): Should we use the passed context or protect users from passing in canceled contexts?
```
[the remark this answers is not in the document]
mikeykhalil: Thanks for your input. Removed the TODO comment and am not using the passed in context. 

I did end up using a new context with a timeout to avoid the potential goroutine leak.

migrations/0019_env_locks_up.sql
```diff
@@ -0,0 +1,14 @@
...
+  PRIMARY KEY(repo, pull_request)
+);
+
+
+CREATE OR REPLACE FUNCTION random_bigint()
+RETURNS bigint AS $$
+BEGIN
+  RETURN floor(random()*(9223372036854775807)); -- between 0 - max bigint
+END;
+$$ LANGUAGE plpgsql;
```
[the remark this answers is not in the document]
mikeykhalil: Correct, with the unique constraint the insert returns an error, which seems reasonable to me.

pkg/persistence/pg.go
mikeykhalil: This operation (now in the `postgres_lock.go` file in the `LockKey` method) is only going to create a Lock key once, the first time a PR/Repo combination receives a webhook event (e.g. PR opened).

After that, we want all other operations to use the same Lock key. 

So if we went with a simple insert, we would be receiving errors from INSERT for all events except the opening of the PR. 

Receiving these errors is fine, but handling that insertion error and then attempting to obtain the value would likely require performing retries of this transaction, no matter what isolation level we use: 
https://www.postgresql.org/docs/current/transaction-iso.html

I'm certainly no postgres expert though, so there could be a simpler way. But I don't think a traditional insert will do the trick in this case.

mikeykhalil: Think we found a way to avoid the table lock. Working through a couple issues with the tests and then will push an update.

pkg/locker/lock_test.go
```diff
@@ -0,0 +1,256 @@
...
+			if tt.tfunc == nil {
+				t.Skip("test func is nil")
+			}
+			tt.tfunc(t, plfunc(t))
+		})
+	}
+}
+
+func testLockAndUnlock(t *testing.T, lp LockProvider) {
+	key := rand.Int63()
```
bkeroackdsc, line 121: Since you're using `math/rand` you'll want to seed the RNG otherwise you'll get the same deterministic sequence of integers every execution.

mikeykhalil: Good catch. Will make a follow up PR to address this.

PR #111, opened 2020-08-25
Fix Acyl Config Test

esterlinej:
Update acyl config test on minikube
 - Add support for latest UI features


cleanshavenalex:
LGTM

cmd/test.go
```diff
@@ -438,17 +437,27 @@ func runUI(dl persistence.DataLayer, repo string) (*http.Server, error) {
 	uilogger := log.New(os.Stderr, "ui server: ", log.LstdFlags)
 	server := &http.Server{}
 	httpapi := api.NewDispatcher(server)
+	tcfg := metahelm.TillerConfig{
+		ServerConnectRetries:    10,
+		ServerConnectRetryDelay: 2 * time.Second,
+	}
+	ci, err := metahelm.NewChartInstallerWithClientsetFromContext(nil, dl, nitromgr.FS, nitromgr.MC, testEnvCfg.k8sCfg.GroupBindings, testEnvCfg.k8sCfg.PrivilegedRepoWhitelist, testEnvCfg.k8sCfg.SecretInjections, tcfg, testEnvCfg.kubeCfgPath, testEnvCfg.kubeCtx)
```
bkeroackdsc, line 444: I'd prefer if you would reuse the existing ChartInstaller instance and pass it into this function, versus instantiating a new one.

bkeroackdsc: Your other option is to convert the `metahelm.Installer` interface type to the underlying concrete struct type here. `ci, ok := nitromgr.CI.(*metahelm.ChartInstaller)`

esterlinej: Perfect, I wasn't a fan of the new instantiation, but didn't want to rearchitect the other methods either.

cmd/test.go
```diff
@@ -438,17 +437,22 @@ func runUI(dl persistence.DataLayer, repo string) (*http.Server, error) {
 	uilogger := log.New(os.Stderr, "ui server: ", log.LstdFlags)
 	server := &http.Server{}
 	httpapi := api.NewDispatcher(server)
+	ci, ok := nitromgr.CI.(*metahelm.ChartInstaller); if !ok {
+		return nil, errors.Errorf("error getting chart installer")
```
bkeroackdsc, line 441: Include the type in the error: `fmt.Errorf("unexpected type for ChartInstaller: %T", nitromgr.CI)`

cmd/test.go
```diff
@@ -438,17 +437,22 @@ func runUI(dl persistence.DataLayer, repo string) (*http.Server, error) {
 	uilogger := log.New(os.Stderr, "ui server: ", log.LstdFlags)
 	server := &http.Server{}
 	httpapi := api.NewDispatcher(server)
+	ci, ok := nitromgr.CI.(*metahelm.ChartInstaller); if !ok {
```
bkeroackdsc, line 440: Add line break instead of semicolon.

PR #112, opened 2020-08-28
Add User Settings

esterlinej:
### User Settings
- Implement a user view in the UI, for user-specific settings/info
- The initial view should be barebones with little or no added functionality. Maybe just a “logout” button that invalidates the session.

-------------------------------------------------------------------------

### Details
- Regarding the user view: pretty much every authenticated web UI has a view/page for user or account settings. Edit profile, change password, log out, etc. 
- Usually symbolized with an icon similar to https://fontawesome.com/icons/user?style=solid
- in most cases (including DSC) the user settings icon is put in the upper right hand corner of the UI
- the complication for Acyl UI is that location (upper right corner) is where the DQA link is (at least for the status page)
- so we need to come up with a location for the "user settings" icon that doesn't conflict and makes intuitive sense
- in terms of the contents of the user setting view, for now probably just a logout link/button is sufficient just as a placeholder (this would invalidate the session)

--------------------------------------------------------------------------

### Updates
- **2020-09-10** Remove Logout, only supply `GitHub User: username` heading in dropdown at this time.
- **2020-09-18** Add `getSessionFromContext` to `statusHandler`
- **2020-09-21** Update `DummySessionUser` to authenticate user, coping session on initial page load for mockui
- **2020-09-21** Modify `defaultBaseTemplateData()` requiring `models.UISession`


esterlinej:
user-settings

pkg/api/api_ui.go
```diff
@@ -88,6 +88,7 @@ type uiapi struct {
 var partials = map[string]string{
 	"header": path.Join("views", "partials", "header.html"),
 	"footer": path.Join("views", "partials", "footer.html"),
+	"user": path.Join("views", "partials", "user.html"),
```
bkeroackdsc, line 91: I'm not sure if this needs to be a separate partial, since it will always be present in the header.

pkg/api/api_ui.go
```diff
@@ -289,13 +290,15 @@ type BaseTemplateData struct {
...
 	RenderEventLink bool
+	DisplaySettings bool
 }
 
 func (api *uiapi) defaultBaseTemplateData() BaseTemplateData {
 	return BaseTemplateData{
 		APIBaseURL:      api.apiBaseURL,
 		Branding:        api.branding,
 		RenderEventLink: false,
+		DisplaySettings: false,
```
bkeroackdsc, line 301: I'm not sure what "DisplaySettings" means.

ui/views/partials/user.html
```diff
@@ -0,0 +1,29 @@
+{{ define "user-settings" }}
```
bkeroackdsc, line 1: I don't think this needs to be a separate partial, since this is only used in one spot.

ui/views/partials/header.html
```diff
@@ -75,53 +75,65 @@
...
+                                role="button"
+                                id="status-link-btn"
+                        >
+                            <small id="status-icon" class="fas status-icon mr-1"></small>
+                            Open DQA
+                        </a>
+                    </form>
+                </div>
+                {{ end }}
+                {{ if .DisplaySettings }}
```
bkeroackdsc, line 129: Since every view is authenticated, I don't think we need this boolean.

ui/assets/user.js
```diff
@@ -0,0 +1,64 @@
...
+    menu.appendChild(h);
+    userSettingsGroup.appendChild(menu);
+    setUserSettingsMenu(userSettingsGroup);
+}
+
+// user gets settings for the current session
+function user() {
+    let req = new XMLHttpRequest();
+
+    req.open('GET', `${apiBaseURL}/v2/user/settings`, true);
```
bkeroackdsc, line 42: The user information is part of the session, so you can use server-side templating to insert this data (we do that elsewhere for GitHub user name) instead of AJAX.

bkeroackdsc (changes requested):

ui/assets/user.js
```diff
@@ -0,0 +1,64 @@
+// setUserSettings replaces the existing user settings group
+function setUserSettingsMenu(userSettingsGroup) {
+    if (userSettingsGroup === null) {
```
mikeykhalil, line 3: Since the `===` is being used here, i think `userSettingsGroup` could be undefined  and this expression would evaluate to false. 

2 possible alternatives:
1. `if (userSettingsGroup == null)` // no type equality check
2. `if (!userSettingsGroup)` // let javascripts truthy/falsy logic handle it

pkg/api/api_ui.go
esterlinej: This is to display the user settings on the page, gating which pages display the settings dropdown.

ui/views/partials/header.html
esterlinej: Sure, though we discussed that the error pages shouldn't display the user settings, we need a way to gate this.

ui/assets/user.js
bkeroackdsc: For example:
https://github.com/dollarshaveclub/acyl/blob/bc372c67572dd4d8ea7d41c05ff00ce081082b88/ui/views/home.html#L48

esterlinej: Okay, the only issue is that we don't have the session for Status page, so we'll need to update that to work

pkg/api/api_ui.go
```diff
@@ -441,7 +452,9 @@ func (ui *uiapi) authenticate(f http.HandlerFunc) http.HandlerFunc {
 					w.WriteHeader(http.StatusInternalServerError)
 					return
 				}
-				f(w, r)
+				uis.Authenticated = true
```
esterlinej, line 455: We may want to add an optional toggle `DummySessionAuthenticated` to control this from the mockui, but for now it's set to true.

pkg/api/api_ui.go
```diff
@@ -674,13 +687,15 @@ func (api *uiapi) homeHandler(w http.ResponseWriter, r *http.Request) {
...
 	}
+	td.BaseTemplateData.RenderUserSettings = true
 	api.render(w, "home", &td)
 }
 
 type envTmplData struct {
 	BaseTemplateData
 	EnvName       string
 	RenderActions bool
+	GitHubUser    string
```
bkeroackdsc, line 698: If this is required for every view, it should be in the base template data.

pkg/api/api_ui.go
```diff
@@ -286,22 +286,25 @@ func (api *uiapi) render(w http.ResponseWriter, name string, td interface{}) {
...
 
 func (api *uiapi) defaultBaseTemplateData() BaseTemplateData {
 	return BaseTemplateData{
-		APIBaseURL:      api.apiBaseURL,
-		Branding:        api.branding,
-		RenderEventLink: false,
+		APIBaseURL:         api.apiBaseURL,
+		Branding:           api.branding,
+		RenderEventLink:    false,
+		RenderUserSettings: false,
```
bkeroackdsc, line 300: This should probably default to true.

pkg/api/api_ui.go
```diff
@@ -286,22 +286,25 @@ func (api *uiapi) render(w http.ResponseWriter, name string, td interface{}) {
...
-	APIBaseURL      string
-	Branding        uiBranding
-	RenderEventLink bool
+	APIBaseURL         string
+	Branding           uiBranding
+	RenderEventLink    bool
+	RenderUserSettings bool
 }
 
 func (api *uiapi) defaultBaseTemplateData() BaseTemplateData {
```
bkeroackdsc, line 295: This could take a ui session and then set the GitHub user field.

pkg/api/api_ui.go
```diff
@@ -286,22 +286,25 @@ func (api *uiapi) render(w http.ResponseWriter, name string, td interface{}) {
 }
 
 type BaseTemplateData struct {
-	APIBaseURL      string
-	Branding        uiBranding
-	RenderEventLink bool
+	APIBaseURL         string
+	Branding           uiBranding
+	RenderEventLink    bool
+	RenderUserSettings bool
```
bkeroackdsc, line 292: Since GitHub user is needed for user settings, that field should be here.

bkeroackdsc (changes requested):

pkg/api/api_ui.go
esterlinej: If we're passing uis to this function, then we can set to true if the user isn't empyt/blank

pkg/api/api_ui.go
```diff
@@ -286,17 +286,23 @@ func (api *uiapi) render(w http.ResponseWriter, name string, td interface{}) {
...
+	Branding                            uiBranding
+	RenderEventLink, RenderUserSettings bool
 }
 
-func (api *uiapi) defaultBaseTemplateData() BaseTemplateData {
-	return BaseTemplateData{
-		APIBaseURL:      api.apiBaseURL,
-		Branding:        api.branding,
-		RenderEventLink: false,
+func (api *uiapi) defaultBaseTemplateData(session *models.UISession) BaseTemplateData {
```
bkeroackdsc, line 294: If you want to pass in a pointer, check for nil prior to using it.

pkg/api/api_ui.go
```diff
@@ -286,17 +286,23 @@ func (api *uiapi) render(w http.ResponseWriter, name string, td interface{}) {
...
+	btd := BaseTemplateData{
+		APIBaseURL:         api.apiBaseURL,
+		Branding:           api.branding,
+		RenderEventLink:    false,
+		RenderUserSettings: false,
 	}
+	if session.GitHubUser != "" {
+		btd.GitHubUser = session.GitHubUser
+		btd.RenderUserSettings = true
+	}
```
bkeroackdsc, line 304: You can move this up to the literal (only if you guarantee that session isn't nil):

```
   GitHubUser: session.GitHubUser,
   RenderUserSettings: session.GitHubUser != "",
```

pkg/api/api_ui.go
```diff
@@ -641,7 +656,7 @@ func (api *uiapi) cacheRenderedAuthErrorPage() {
 		return
 	}
 	b := bytes.Buffer{}
-	if err := api.views["auth_error"].Execute(&b, api.defaultBaseTemplateData()); err != nil {
+	if err := api.views["auth_error"].Execute(&b, api.defaultBaseTemplateData(&models.UISession{})); err != nil {
```
bkeroackdsc, line 659: If `defaultBaseTemplateData()` was changed to handle a nil session, you could avoid unnecessarily allocating a new struct here.

pkg/api/api_ui.go
```diff
@@ -650,7 +665,7 @@ func (api *uiapi) cacheRenderedAuthErrorPage() {
 
 func (api *uiapi) authErrorHandler(w http.ResponseWriter, r *http.Request) {
 	if api.reload || len(api.oauth.errorPage) == 0 {
-		api.render(w, "auth_error", api.defaultBaseTemplateData())
+		api.render(w, "auth_error", api.defaultBaseTemplateData(&models.UISession{}))
```
bkeroackdsc, line 668: Pass nil if you fix the method to handle it properly.

pkg/api/api_ui.go
```diff
@@ -742,5 +755,5 @@ func (api *uiapi) envHandler(w http.ResponseWriter, r *http.Request) {
 }
 
 func (api *uiapi) deniedHandler(w http.ResponseWriter, r *http.Request) {
-	api.render(w, "denied", api.defaultBaseTemplateData())
+	api.render(w, "denied", api.defaultBaseTemplateData(&models.UISession{}))
```
bkeroackdsc, line 758: Same here.

bkeroackdsc (changes requested):

bkeroackdsc (approved):

PR #113, opened 2020-09-01
locker: print notification payload for debugging

mikeykhalil:
* Fixes issue where LockKey was just returning default value for int64 and not reporting errors properly
* Fixes issues with tests to determine if errors are not being reported and if we see a default value for int64. We could instead switch the return type of LockKey to `(*int64, error)` to differentiate between nil and the default value, but that seems a bit clunky to use.

### DIFF
--- a/pkg/locker/locker.go
+++ b/pkg/locker/locker.go
@@ -184,6 +184,7 @@ func (p *PreemptiveLocker) Lock(ctx context.Context) (ch <-chan NotificationPayl
 
 	p.key = key
 	span.SetTag("lock_key", p.key)
+	p.log(ctx, "locking key: %d for repo: %s, pr: %d", p.key, p.repo, p.pr)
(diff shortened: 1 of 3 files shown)


pkg/locker/locker.go
```diff
@@ -202,8 +202,8 @@ func (p *PreemptiveLocker) Lock(ctx context.Context) (ch <-chan NotificationPayl
 
 	// Wait for the specified LockDelay before returning
 	select {
-	case <-ch:
-		return nil, ErrLockPreempted
+	case np := <-ch:
+		return nil, errors.Wrap(ErrLockPreempted, np.Message)
```
bkeroackdsc, line 206: I would add documentation to highlight that the error returned by this method will not equal `ErrLockPreempted` due to this wrap. Clients will need to unwrap the error to check for `ErrLockPreempted`.

Consider adding a `IsPreemptedError(err error) bool` function for clients as a shorthand.

PR #114, opened 2020-09-02
locker: fix issue when preempted while waiting for lock

mikeykhalil:
Noticed an issue with the PreemptiveLocker when the lock was preempted while waiting for the lock delay. 

Since users will typically only do a `defer pl.Release(ctx)` if no error is returned from `pl.Lock(ctx)`, I think it makes sense to Release on behalf of the user in this scenario. 

Although, thinking more about it, I think it would probably just make sense to not even return this error, but to instead just return the channel. The user will then use the same logic to determine when to preempt as if it was done later on in the operation.

### DIFF
--- a/pkg/locker/locker.go
+++ b/pkg/locker/locker.go
@@ -202,7 +202,6 @@ func (p *PreemptiveLocker) Lock(ctx context.Context) (ch <-chan NotificationPayl
 	}
 
 	select {
-	// TODO (mk): Might be simpler to just let the user recognize that this operation has already been preempted, since they will be reading from the channel anyways.
 	case np := <-ch:
 		releaseCtx, cancel := context.WithTimeout(context.Background(), p.conf.lockWait)
 		releaseErr := p.Release(releaseCtx)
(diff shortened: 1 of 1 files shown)


pkg/locker/locker.go
```diff
@@ -201,10 +201,17 @@ func (p *PreemptiveLocker) Lock(ctx context.Context) (ch <-chan NotificationPayl
 		return nil, errors.Wrap(err, "unable to lock")
 	}
 
-	// Wait for the specified LockDelay before returning
 	select {
+	// TODO (mk): Might be simpler to just let the user recognize that this operation has already been preempted, since they will be reading from the channel anyways.
```
mikeykhalil, line 205: For context, this select statement is basically providing an error to the caller of the `Lock` method in the case that the lock was preempted during the lock delay portion. 

However, we return the same channel at the end of this method anyways, so the user could also make the decision to release the lock by receiving from the same channel, and avoid this duplicate logic.

bkeroackdsc: I'm ok with the logic as-is. Semantically it's not really correct to return a nil error if we already know the lock has been preempted. I like that the method doesn't wait for the full lock wait period if it gets preempted.

mikeykhalil: > Semantically it's not really correct to return a nil error if we already know the lock has been preempted. I like that the method doesn't wait for the full lock wait period if it gets preempted.

Good points. Removed the TODO comment.

bkeroackdsc (approved):

Issue #90, opened 2020-05-21
Timeout doesn't update multiple things


bkeroackdsc:
When an event timeout occurs, several things don't happen that should happen:

* QAEnvironment status field is not updated from "Pending" (Spawned/Updating).
* GitHub commit status is not set to to failed (stays as Pending)
* Various fields within the EventLog Status object are not updated to failed (whatever images or charts that were pending when the timeout happened stay pending forever).

bkeroackdsc:
Fixed by #115

PR #115, opened 2020-09-02
Set status timeout

mikeykhalil:
We had previously made a change where we updated callers of `SetQAEnvironmentStatus` to pass in the background context. This was in response to an issue where the status would not be updated when operations failed or were preempted, since the context would often be canceled and prevent that status call from actually behaving as expected.

However, I just saw an issue where the `SetQAEnvironmentStatus` function hung, and now recognize that perhaps we should have a proper timeout set for this call. 

The particular issue I saw resulted in:
1. The environment being stuck in the status of "Pending"
2. The lock never getting released since `Release(ctx)` was never called. 

So this prevented subsequent events (e.g. closing the PR) from being able to obtain the lock and clean up the environment. 

My proposal is to just add a timeout from within the function itself. This way, callers can still have the convenience of just passing in a background context (if they would like), without having to worry about the operation potentially hanging forever.

### DIFF
--- a/pkg/persistence/pg.go
+++ b/pkg/persistence/pg.go
@@ -284,7 +284,9 @@ func (p *PGLayer) GetQAEnvironmentsByUser(ctx context.Context, user string) ([]Q
 }
 
 // SetQAEnvironmentStatus sets a specific QAEnvironment's status.
-// Note that this will bail if the context was canceled. It is often recommended to pass the background context to this function.
+// This method is unique in that callers often call this in cases where a root operation context is likely to be canceled.
+// Because of this, it often makes sense to pass the background context, or a newly created child of the background context, to this function.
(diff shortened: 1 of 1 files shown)


bkeroackdsc:
@mikeykhalil Interesting. Could this be related to #90? We should investigate why the postgres query is hanging.

mikeykhalil:
> @mikeykhalil Interesting. Could this be related to #90? We should investigate why the postgres query is hanging.

Yes, this does seem to be related to #90. Now that we use the postgres locker, it seems this issue is a bit more painful since there is no server side release of the lock (apart from when the connection is terminated which could theoretically be when the pod is recycled). 

I figure having a timeout is a good start, but I 100% agree that we should investigate why this query is failing before we can mark the issue as resolved.

bkeroackdsc (approved):

PR #118, opened 2020-09-15
Create unknown status prior to actions

esterlinej:


pkg/eventlogger/status_test.go
```diff
@@ -273,3 +274,37 @@ func TestSetCompletedStatus(t *testing.T) {
 		t.Fatalf("unexpected status: %v", status)
 	}
 }
+
+func TestSetNewStatusUnknown(t *testing.T) {
```
mikeykhalil, line 278: Think I misunderstood which package this test was located in and the full contents of the tests when reading it in Slack.

I'm not sure if this test alone does enough to prevent a regression of this issue or ensure it has been fixed. 

Don't have the code on hand at the moment, but perhaps we can validate the event log status in an API handler test or in the environment creation test?

esterlinej: I added a secondary test for env_test, hopefully the combination of both tests will catch any regression

pkg/nitro/env/env_test.go
```diff
@@ -1550,3 +1550,74 @@ func TestSetGithubCommitStatus(t *testing.T) {
...
+			Revision: rrd.SourceSHA,
+			Started: el2.Config.Started,
+			Completed: el2.Config.Completed,
+		},
+	}
+	if !cmp.Equal(el2, exp) {
+		t.Fatalf("expected comparison to match:\nRsp: %+v\nExp: %+v", el2, exp)
+	}
+
+	env, err := m.generateNewEnv(context.Background(), &rrd)
```
mikeykhalil, line 1604: What if we call `m.create` here to better mimic what would happen in the API handler and to test out multiple calls to `SetNewStatus`?

For instance, I think the behavior that this PR changes is that it calls `SetNewStatus` almost immediately within the handler, and then it's called again in functions like `m.create`. So we should want this test to basically verify that

1. The event log is able to be set to an unknown event status initially
2. The event log status gets updated to `DoneStatus` after m.create finishes.

mikeykhalil (changes requested):

esterlinej:
Screen Shot 2020-09-16 at 5 43 25 PM

pkg/nitro/env/env_test.go
```diff
@@ -1550,3 +1550,163 @@ func TestSetGithubCommitStatus(t *testing.T) {
...
+		NG:               &namegen.FakeNameGenerator{},
+		FS:               memfs.New(),
+		MG:               fg,
+		RC:               frc,
+		CI:               ci,
+		GlobalLimit:      0,
+		OperationTimeout: 0,
+	}
+	envName, err := m.create(ctx, &rrd)
+	if err != nil {
```
mikeykhalil, line 1687: I think we can get rid of lots of these validations, since `TestCreate` performs these same validations. 

What we need to verify here is that the event log status gets updated after the call to `create`. I believe this can be retrieved by calling `dl. GetEventStatus`

esterlinej: Simple enough, will update

pkg/nitro/env/env_test.go
```diff
@@ -1550,3 +1550,163 @@ func TestSetGithubCommitStatus(t *testing.T) {
...
+		"foo-bar":    meta.ChartLocation{ChartPath: "/tmp/foo/bar", VarFilePath: "/tmp/foo/bar/vars.yml"},
+		"foo-qwerty": meta.ChartLocation{ChartPath: "/tmp/foo/qwerty", VarFilePath: "/tmp/foo/qwerty/vars.yml"},
+		"foo-mysql":  meta.ChartLocation{ChartPath: "/tmp/foo/mysql", VarFilePath: "/tmp/foo/mysql/vars.yml"},
+	}
+	bm := map[string][]ghclient.BranchInfo{
+		"foo/bar":    []ghclient.BranchInfo{ghclient.BranchInfo{Name: "feature-spam"}, ghclient.BranchInfo{Name: "release"}},
+		"foo/qwerty": []ghclient.BranchInfo{ghclient.BranchInfo{Name: "feature-spam"}, ghclient.BranchInfo{Name: "release"}},
+		"foo/mysql":  []ghclient.BranchInfo{ghclient.BranchInfo{Name: "feature-spam"}, ghclient.BranchInfo{Name: "release"}, ghclient.BranchInfo{Name: "master"}},
+	}
+	//cir := map[string]error{
```
mikeykhalil, line 1612: if not needed, perhaps we can just delete this?

esterlinej: Yeah, meant to do that

mikeykhalil (approved):

bkeroackdsc (approved):

Issue #116, opened 2020-09-09
Locker: Lock may be incorrectly held until the postgres connection terminates


mikeykhalil:
**Problem Description**
For postgres session-level advisory locks, there are 2 ways for the lock to get released:
1. The same session calls unlock
2. The session ends (connection is terminated)

Acyl uses advisory locks to lock on a repo/pr combination.

With Acyl's goroutine model, there are 3 goroutines that are important to understand that relate to locking:
1. The goroutine that is in charge of creating the environment (we will call Goroutine A)
2. The goroutine managing the state of the postgres lock (we will call Goroutine B)
3. The goroutine that listens for preemption events and cancels the root context for the environment event (we will call Goroutine C)

Goroutine B listens for notifications from Postgres, and relays these notifications over a channel to Goroutine C. The message sent over the channel informs Goroutine C to cancel the context. It is then expected that Goroutine A would respect the context being canceled, exit as soon as possible, and then would eventually call the deferred function which releases the lock. However, there aren't any strong guarantees that Goroutine A will call that deferred function in a reasonable amount of time, as it performs a complex series of steps to update the environment.

This can leave the lock in a locked state indefinitely, and forces an administrator to have to close that postgres connection somehow in order for the lock to get unlocked. 

**Proposed Fix**
There should be a timeout at which point the lock is released automatically. Since we have a global timeout of 30.5 minutes, we could propagate this timeout to the lock itself to ensure that the lock is released and the connection ends no matter what happens during the creation of the environment with Goroutine A. Goroutine C could unlock the lock itself and ensure that new Goroutines can obtain the lock.

PR #119, opened 2020-09-15
fix: issue 116, create safeguards to release lock

mikeykhalil:
Fixes https://github.com/dollarshaveclub/acyl/issues/116

**Summary**
This introduces 2 new fields to the LockProvider implementations. 
1. `MaxLockDuration`: After this duration, the goroutine that manages the state of the Postgres Lock will force the lock to be released. This prevents locks from being held indefinitely, in the event that the holder never calls `Unlock(ctx)`
2. `PreemptionTimeout`: After the postgres lock goroutine receives a notification that the lock should be released, it will immediately send a message over the `preempt` channel returned from the `Lock` method. After the `PreemptionTimeout` duration, it will force the lock to Unlock, in the event the holder doesn't respond. 

**Specifics** 
* Ensure and document that `Unlock` is idempotent, since it might be called multiple times now.
* Create a single factory,`NewLockProvider`, for creating both Fake and Postgres LockProviders. This approach was chosen so that the logic to set default values and perform validation on passed configuration is done in a single function. 
* Reduced the `defaultLockWait` value from 30.5 minutes to 20 seconds, since returning the error more promptly could prevent users from waiting very long in the event there is an issue. Now that we have a `MaxLockDuration` value, this will be the value which protects against leaking the Goroutine or having that particular Goroutine hang.
* Listens for notifications before acquiring the lock. This prevents the lock from potentially missing any notifications while waiting to obtain the lock from Postgres. 
* Updated tests and usages of LockProviders to use the `NewLockProvider` function
* Updated `lockingOperation` method in `env.go` to ensure that it always returns when a context is canceled. This should prevent updates from getting stuck in a pending state and ensure that the lock is explicitly closed.

### DIFF
--- a/cmd/integration.go
+++ b/cmd/integration.go
@@ -281,13 +281,11 @@ func setupNitro(dl persistence.DataLayer, useGHToken bool) (spawner.EnvironmentS
 	rc := ghclient.NewGitHubClient(ghtkn)
 	ng := &namegen.FakeNameGenerator{Unique: true}
 	mc := &metrics.FakeCollector{}
-	lp, err := locker.NewLockProvider(locker.FakeLockProviderKind, locker.WithLockWait(1*time.Second))
+	plf, err := locker.NewFakePreemptiveLockerFactory([]locker.LockProviderOption{
+		locker.WithLockTimeout(1 * time.Second),
+	})
 	if err != nil {
-		return nil, nil, errors.Wrap(err, "error creating lock provider")
-	}
-	plf, err := locker.NewPreemptiveLockerFactory(lp)
-	if err != nil {
-		return nil, nil, errors.Wrap(err, "error creating preemptive locker factory")
+		return nil, nil, errors.Wrap(err, "unable to create new fake preeemptive locker factory")
 	}
 	nf := func(lf func(string, ...interface{}), notifications models.Notifications, user string) notifier.Router {
 		sb := &notifier.SlackBackend{
--- a/cmd/integration.go
+++ b/cmd/integration.go
@@ -281,7 +281,11 @@ func setupNitro(dl persistence.DataLayer, useGHToken bool) (spawner.EnvironmentS
 	rc := ghclient.NewGitHubClient(ghtkn)
 	ng := &namegen.FakeNameGenerator{Unique: true}
 	mc := &metrics.FakeCollector{}
-	plf, err := locker.NewPreemptiveLockerFactory(locker.NewFakeLockProvider())
+	lp, err := locker.NewLockProvider(locker.FakeLockProviderKind, locker.WithLockWait(1*time.Second))
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "error creating lock provider")
+	}
+	plf, err := locker.NewPreemptiveLockerFactory(lp)
 	if err != nil {
 		return nil, nil, errors.Wrap(err, "error creating preemptive locker factory")
 	}
--- a/pkg/locker/fake_lock.go
+++ b/pkg/locker/fake_lock.go
@@ -88,15 +88,9 @@ func (fls *fakeLockStore) notify(ctx context.Context, key int64, id uuid.UUID) e
 			ID:      lock.id,
 			Message: lock.event,
 		}
-		go fls.handleNotification(lock, np)
+		go lock.handleNotification(np)
 	}
 	return nil
-}
-
-func (fls *fakeLockStore) handleNotification(lock *FakePreemptableLock, np NotificationPayload) {
-	go func() { lock.preempt <- np }()
-	time.Sleep(lock.conf.forcePreemption)
-	lock.Unlock(context.Background())
 }
 
 var _ LockProvider = &FakeLockProvider{}
@@ -136,7 +130,7 @@ func (flp *FakeLockProvider) New(ctx context.Context, key int64, event string) (
 	}
 	l := &FakePreemptableLock{
 		id:      id,
-		preempt: make(chan NotificationPayload, 1000),
+		preempt: make(chan NotificationPayload),
 		store:   flp.store,
 		event:   event,
 		key:     key,
@@ -158,15 +152,20 @@ type FakePreemptableLock struct {
 }
 
 func (fpl *FakePreemptableLock) Lock(ctx context.Context) (<-chan NotificationPayload, error) {
-	lockCtx, cancel := context.WithTimeout(ctx, fpl.conf.lockWait)
+	lockCtx, cancel := context.WithTimeout(ctx, fpl.conf.lockTimeout)
 	defer cancel()
 	err := fpl.store.lock(lockCtx, fpl.key)
 	if err != nil {
 		return nil, fmt.Errorf("unable to lock fake preemptable lock: %v", err)
 	}
 	go func() {
-		time.Sleep(fpl.conf.forceUnlock)
-		fpl.store.unlock(context.Background(), fpl.key, fpl.id)
+		time.Sleep(fpl.conf.maxLockDuration)
+		np := NotificationPayload{
+			ID:      fpl.id,
+			Message: "reached max lock duration",
+			LockKey: fpl.key,
+		}
+		fpl.handleNotification(np)
 	}()
 
 	return fpl.preempt, nil
@@ -178,4 +177,13 @@ func (fpl *FakePreemptableLock) Unlock(ctx context.Context) error {
 
 func (fpl *FakePreemptableLock) Notify(ctx context.Context) error {
 	return fpl.store.notify(ctx, fpl.key, fpl.id)
+}
+
+func (fpl *FakePreemptableLock) handleNotification(np NotificationPayload) {
+	select {
+	case fpl.preempt <- np:
+		return
+	case <-time.After(fpl.conf.preemptionTimeout):
+		fpl.Unlock(context.Background())
+	}
 }
--- a/pkg/locker/fake_lock.go
+++ b/pkg/locker/fake_lock.go
@@ -83,16 +83,20 @@ func (fls *fakeLockStore) notify(ctx context.Context, key int64, id uuid.UUID) e
 		if lock.id == id {
 			continue
 		}
-		go func(l *FakePreemptableLock) {
-			l.preempt <- NotificationPayload{
-				LockKey: key,
-				ID:      l.id,
-				Message: l.event,
-			}
-		}(lock)
-
+		np := NotificationPayload{
+			LockKey: key,
+			ID:      lock.id,
+			Message: lock.event,
+		}
+		go fls.handleNotification(lock, np)
 	}
 	return nil
+}
+
+func (fls *fakeLockStore) handleNotification(lock *FakePreemptableLock, np NotificationPayload) {
+	go func() { lock.preempt <- np }()
+	time.Sleep(lock.conf.forcePreemption)
+	lock.Unlock(context.Background())
 }
 
 var _ LockProvider = &FakeLockProvider{}
@@ -100,14 +104,16 @@ var _ LockProvider = &FakeLockProvider{}
 // FakeLockProvider can serve as a lock provider with no external dependencies
 type FakeLockProvider struct {
 	store *fakeLockStore
+	conf  LockProviderConfig
 }
 
-func NewFakeLockProvider() *FakeLockProvider {
+func newFakeLockProvider(conf LockProviderConfig) *FakeLockProvider {
 	return &FakeLockProvider{
 		store: &fakeLockStore{
 			locks:        make(map[int64]*fakeLockStoreEntry),
 			envToLockKey: make(map[string]int64),
 		},
+		conf: conf,
 	}
 }
 
@@ -134,6 +140,7 @@ func (flp *FakeLockProvider) New(ctx context.Context, key int64, event string) (
 		store:   flp.store,
 		event:   event,
 		key:     key,
+		conf:    flp.conf,
 	}
 	flp.store.store(key, l)
 	return l, nil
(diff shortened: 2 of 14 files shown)


pkg/locker/locker.go
```diff
@@ -12,17 +13,109 @@ import (
...
-	defaultLockDelay    = 10 * time.Second
+	defaultForcePreemption = 10 * time.Second
+	defaultLockWait        = 20 * time.Second
+	defaultLockDelay       = 10 * time.Second
+	defaultForceUnlock     = 30*time.Minute + 30*time.Second // global async timeout is 30 minutes
 )
 
+// WithForceUnlock sets the duration for which the lock will wait before forceably unlocking the lock.
+// This can be used to protect against the holder of the lock getting blocked or not being able to unlock the lock.
+func WithForceUnlock(duration time.Duration) LockProviderOption {
```
bkeroackdsc, line 28: A better name would be "UnlockTimeout" or "MaximumLockDuration".

pkg/locker/locker.go
```diff
@@ -12,17 +13,109 @@ import (
...
+	return func(config *LockProviderConfig) {
+		config.forceUnlock = duration
+	}
+}
+
+// WithForcePreemption sets the duration for which the lock will wait for the holder to respect a notification and release the lock.
+// If a notification is sent to the lock, the lock will be released after this duration no matter what the holder does.
+// This provides a mechanism for releasing the lock quickly in scenarios where the holder of the lock may be blocked indefinitely.
+func WithForcePreemption(duration time.Duration) LockProviderOption {
```
bkeroackdsc, line 37: "PreemptionTimeout" is a better name for this.

pkg/locker/locker.go
```diff
@@ -12,17 +13,109 @@ import (
...
+	return func(config *LockProviderConfig) {
+		config.forcePreemption = duration
+	}
+}
+
+// WithLockWait sets the duration for which the lock will block while attempting to lock, before it gives up and returns an error.
+func WithLockWait(lockWait time.Duration) LockProviderOption {
```
bkeroackdsc, line 44: "LockTimeout" is a better name.

cmd/integration.go
```diff
@@ -281,7 +281,11 @@ func setupNitro(dl persistence.DataLayer, useGHToken bool) (spawner.EnvironmentS
 	rc := ghclient.NewGitHubClient(ghtkn)
 	ng := &namegen.FakeNameGenerator{Unique: true}
 	mc := &metrics.FakeCollector{}
-	plf, err := locker.NewPreemptiveLockerFactory(locker.NewFakeLockProvider())
+	lp, err := locker.NewLockProvider(locker.FakeLockProviderKind, locker.WithLockWait(1*time.Second))
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "error creating lock provider")
+	}
+	plf, err := locker.NewPreemptiveLockerFactory(lp)
```
bkeroackdsc, line 288: If these always follow each other (`locker.NewLockProvider()` followed by `locker.NewPreemptiveLockerFactory()`) then you should combine them into a single function.

pkg/locker/fake_lock.go
```diff
@@ -83,31 +83,37 @@ func (fls *fakeLockStore) notify(ctx context.Context, key int64, id uuid.UUID) e
...
+		}
+		go fls.handleNotification(lock, np)
 	}
 	return nil
 }
 
+func (fls *fakeLockStore) handleNotification(lock *FakePreemptableLock, np NotificationPayload) {
+	go func() { lock.preempt <- np }()
+	time.Sleep(lock.conf.forcePreemption)
+	lock.Unlock(context.Background())
```
bkeroackdsc, line 99: If I'm understanding this correctly, this should probably be a `select` with one case that sends on `lock.preempt` and another that waits on a timer that does the forced unlock once the forcePreemption duration is elapsed.

pkg/locker/fake_lock.go
```diff
@@ -147,15 +154,21 @@ type FakePreemptableLock struct {
...
+	lockCtx, cancel := context.WithTimeout(ctx, fpl.conf.lockWait)
 	defer cancel()
 	err := fpl.store.lock(lockCtx, fpl.key)
 	if err != nil {
 		return nil, fmt.Errorf("unable to lock fake preemptable lock: %v", err)
 	}
+	go func() {
+		time.Sleep(fpl.conf.forceUnlock)
+		fpl.store.unlock(context.Background(), fpl.key, fpl.id)
+	}()
```
bkeroackdsc, line 170: This seems incorrect. It will always stay alive for the full force unlock duration, while it should return when the lock is otherwise preempted or unlocked normally.

pkg/locker/locker.go
```diff
@@ -12,17 +13,109 @@ import (
...
+	}
+}
+
+// WithPostgresBackend allows for a distributed lock provider backed by Postgres.
+// APM traces can also be enabled with the configured service name.
+func WithPostgresBackend(postgresURI string, enableTracing bool, apmServiceName string) LockProviderOption {
+	return func(config *LockProviderConfig) {
+		config.postgresURI = postgresURI
+		config.apmServiceName = apmServiceName
+		config.enableTracing = enableTracing
```
bkeroackdsc, line 56: Possibly redundant. A non-empty APM service name implies tracing should be enabled.

pkg/locker/locker.go
```diff
@@ -12,17 +13,109 @@ import (
...
+// LockProviderKind enumerates the different LockProvider implementations
+type LockProviderKind int
+
+const (
+	// PostgresLockProviderKind represents an implementation of the LockProvider interface that uses Postgres as the underlying store
+	PostgresLockProviderKind LockProviderKind = iota
+
+	// FakeLockProviderKind represents an implementation of the LockProvider interfaces that uses an in-memory store, suitable for tests
+	FakeLockProviderKind
+)
```
bkeroackdsc, line 85: When implementing enums, the first value should almost always be "Unknown"/unset to indicate the zero value.

pkg/locker/postgres_lock.go
```diff
@@ -186,23 +189,29 @@ func (pl *PostgresLock) handleEvents(ctx context.Context, listener *pq.Listener)
...
 				return
 			}
 		}
 	}
 }
 
+func (pl *PostgresLock) handleNotification(np NotificationPayload) {
+	go func() { pl.preempted <- np }()
+	time.Sleep(pl.conf.forcePreemption)
+	pl.Unlock(context.Background())
```
bkeroackdsc, line 212: Same comment here as for the same function in the fake. It seems like you'd want a `select` here.

pkg/locker/postgres_lock.go
```diff
@@ -269,6 +270,19 @@ func (pl *PostgresLock) Lock(ctx context.Context, lockWait time.Duration) (<-cha
 		pl.handleEvents(ctx, pl.listener)
 	}()
 
+	go func() {
+		time.Sleep(pl.conf.forceUnlock)
+		pl.Unlock(context.Background())
+	}()
```
bkeroackdsc, line 276: Similar issue here as the fake comment above. This goroutine should exit when the lock is normally unlocked or preempted.

bkeroackdsc (changes requested):

pkg/locker/locker.go
mikeykhalil: Fixed here: https://github.com/dollarshaveclub/acyl/pull/119/commits/74e5377c4bc571c1346bdd0a28b81e7cae22c465#diff-3491d15067d4403add4736c7915432e1R26

pkg/locker/locker.go
mikeykhalil: Fixed here: https://github.com/dollarshaveclub/acyl/pull/119/commits/74e5377c4bc571c1346bdd0a28b81e7cae22c465#diff-3491d15067d4403add4736c7915432e1R34

pkg/locker/locker.go
mikeykhalil: Fixed here: https://github.com/dollarshaveclub/acyl/pull/119/commits/74e5377c4bc571c1346bdd0a28b81e7cae22c465#diff-3491d15067d4403add4736c7915432e1R44

cmd/integration.go
mikeykhalil: Added 2 new helper functions, one for each kind of lock provider:

https://github.com/dollarshaveclub/acyl/pull/119/commits/74e5377c4bc571c1346bdd0a28b81e7cae22c465#diff-3491d15067d4403add4736c7915432e1R120

Little awkward because there are 2 sets of functional arguments, but I do think it simplifies things for the caller.

pkg/locker/locker.go
mikeykhalil: True, added that change here: https://github.com/dollarshaveclub/acyl/pull/119/commits/74e5377c4bc571c1346bdd0a28b81e7cae22c465#diff-3491d15067d4403add4736c7915432e1R102

pkg/locker/postgres_lock.go
mikeykhalil: I like this idea a lot and implemented it [here](https://github.com/dollarshaveclub/acyl/pull/119/commits/74e5377c4bc571c1346bdd0a28b81e7cae22c465#diff-bce65b51266068615a521ee5543292f9R210).

While this logic is sound, the way we read from this channel in `env.go` still does not guarantee that the lock will ever get unlocked. This is because we actually depend on a deferred function call from a different goroutine, that does not read from the `preempt` channel directly [here](https://github.com/dollarshaveclub/acyl/blob/master/pkg/nitro/env/env.go#L228).


I have one idea i'm going to try out though to modify how we actually use the preemptive locker in `env.go`, and will post up here when that is ready for review. 

---
As a backup plan though, I think the original logic works as this implementation of Unlock is idempotent. We would be making extraneous calls though.

mikeykhalil: Okay, made the change to env.go [here](https://github.com/dollarshaveclub/acyl/pull/119/files#diff-1ea1b07272feac8de974c49bbec91702R246).

Since the passed in function to `lockingOperation` can block indefinitely, I decided to add some protection to at least ensure the root goroutine ends. This way, even if the blocking operation fails, the goroutine which called `lockingOperation` is able to complete and specifically call the deferred function which unlocks the lock.

pkg/locker/locker.go
mikeykhalil: Fixed here: https://github.com/dollarshaveclub/acyl/pull/119/files#diff-3491d15067d4403add4736c7915432e1R80

pkg/locker/postgres_lock.go
mikeykhalil: Ended up removing this logic, in favor of using a [context](https://github.com/dollarshaveclub/acyl/pull/119/files#diff-bce65b51266068615a521ee5543292f9R160
) with a timeout period within the goroutine that manages `handleEvents`. In the event the context is canceled (e.g. after the maxLockDuration), it will call `handleNotification(np)`. 

Handle notifications perform this logic and will exit normally if preempted.

pkg/locker/fake_lock.go
mikeykhalil: Resolved by making use of the newly created [handleNotification](https://github.com/dollarshaveclub/acyl/pull/119/files#diff-e2a2c0cf8eb3d6bbeabf2693327aaa17R182) method

pkg/locker/fake_lock.go
mikeykhalil: Resolved by making use of the newly created [handleNotification](https://github.com/dollarshaveclub/acyl/pull/119/files#diff-e2a2c0cf8eb3d6bbeabf2693327aaa17R182) method

testing/localsecrets/main.go
```diff
@@ -130,8 +130,8 @@ func injectFuranSecrets(kc *kubernetes.Clientset, ns string) {
 		ferr("error unmarshaling configmap json: %v", err)
 	}
 
-	secrets["secret/production/furan/github/token"] = getGitHubToken()
-	secrets["secret/production/furan/dockercfg"] = getFuranDockerCfg()
+	secrets["production/furan/github/token"] = getGitHubToken()
```
mikeykhalil, line 133: This change was required due to an update to one of the base docker images being used.

pkg/locker/fake_lock.go
```diff
@@ -166,3 +178,12 @@ func (fpl *FakePreemptableLock) Unlock(ctx context.Context) error {
 func (fpl *FakePreemptableLock) Notify(ctx context.Context) error {
 	return fpl.store.notify(ctx, fpl.key, fpl.id)
 }
+
+func (fpl *FakePreemptableLock) handleNotification(np NotificationPayload) {
+	select {
+	case fpl.preempt <- np:
+		return
+	case <-time.After(fpl.conf.preemptionTimeout):
```
bkeroackdsc, line 186: I probably would put a debug log here, something like `preemptable lock: <lock key>: preemption timeout reached, unlocking: 30m`.

pkg/locker/locker.go
bkeroackdsc: The point is to remove the redundant config option, not to set it automatically.

pkg/nitro/env/env.go
```diff
@@ -243,7 +243,18 @@ func (m *Manager) lockingOperation(ctx context.Context, repo string, pr uint, f
...
+	ch := make(chan error)
+	go func() {
+		ch <- f(ctx)
+	}()
+	select {
+	case opErr := <-ch:
+		err = opErr
+	case <-ctx.Done():
+		err = ctx.Err()
+	}
```
bkeroackdsc, line 257: You could create a derived context with a set timeout here as well, if you want additional assurance that the input function will be signaled to abort at some point.

However we can never guarantee that the function will respect the context, so the logic you added is a good fallback.

bkeroackdsc (approved):

pkg/locker/fake_lock.go
bkeroackdsc: Not just the fake, also in the Postgres implementation.

pkg/nitro/env/env.go
mikeykhalil: Yeah, I think the problem is that the passed in functions don't always respect the context, so it's important for this method to make sure the context cancelation is handled properly. 

I think there are some reliable mechanisms in place that ensure the context itself will get canceled at some point, including the Manager's `OperationTimeout` and the PostgresLock's `MaxLockDuration`.

esterlinej (approved):
LGTM, but defer to Ben for greater context

pkg/locker/fake_lock.go
mikeykhalil: [Fake implementation fix](https://github.com/dollarshaveclub/acyl/pull/119/commits/0ea4fe0201e70de4d5622ab43612fe63d9c81c7d#diff-e2a2c0cf8eb3d6bbeabf2693327aaa17R188)

[Postgres implementation fix](https://github.com/dollarshaveclub/acyl/pull/119/commits/0ea4fe0201e70de4d5622ab43612fe63d9c81c7d#diff-bce65b51266068615a521ee5543292f9R214url)

pkg/locker/locker.go
mikeykhalil: Fixed here: https://github.com/dollarshaveclub/acyl/pull/119/commits/0ea4fe0201e70de4d5622ab43612fe63d9c81c7d

PR #120, opened 2020-09-24
add swagger

esterlinej:
### Swagger Docs
- Implement generated API documentation (Swagger, etc) and host it somewhere.
- It would make most sense to host in the UI somewhere (access-controlled)
### Update
- drop go-swagger for simple conversion with  `bootprint-openapi` yaml -> html.


esterlinej:
Screen Shot 2020-09-28 at 2 28 50 PM
Screen Shot 2020-09-28 at 2 29 05 PM
Screen Shot 2020-09-28 at 2 28 58 PM

doc/swagger/swagger.yml
```diff
@@ -0,0 +1,792 @@
...
+      operationId: getEnvs
+      parameters:
+        - name: API-Key
+          in: header
+          description: API key or aborts with Unauthorized
+          required: true
+          type: string
+        - name: full_details
+          in: query
+          description: returns full list of details as defined by the `v0QAEnvironment` struct for each environment
```
mikeykhalil, line 51: Think it might make sense to reference a swagger definition as opposed to a go struct type? Not sure of the best way to accomplish that though.

doc/swagger/swagger.yml
```diff
@@ -0,0 +1,792 @@
...
+          required: false
+          type: string
+      responses:
+        200:
+          description: returns list of environments
+          schema:
+            $ref: '#/definitions/Envs'
```
mikeykhalil, line 58: When `full_details == true`, this is returned. 

However, when `full_details == false`, it seems just a `[]string` is returned, which contains all the env names. 

Is there a way to encode this information in this swagger file?

doc/swagger/swagger.yml
```diff
@@ -0,0 +1,792 @@
...
+        500:
+          schema:
+            $ref: '#/definitions/ServerError'
+  /envs/:
```
mikeykhalil, line 62: I'm not exactly sure why there are 2 separate endpoints, `/envs` and `/envs/`, within Acyl. If we are going to document both, then there needs to be some additional detail provided in this document as to why both exists and when to use which one. 

After a quick glance at the code, it seems both of these endpoints do the same thing. In which case, perhaps we just document a single endpoint and mention in the description that a trailing slash is accepted/optional.

doc/swagger/swagger.yml
```diff
@@ -0,0 +1,792 @@
...
+            $ref: '#/definitions/ClientError'
+        404:
+          description: not found
+          schema:
+            $ref: '#/definitions/ClientError'
+        500:
+          description: internal server error
+          schema:
+            $ref: '#/definitions/ServerError'
+definitions:
```
mikeykhalil, line 734: Within these definitions, we want to provide definitions of the JSON encoded objects we return to clients. 

HTTP status codes should be removed from this section, unless they are included within the JSON encoded object we return to the client. 

Think this is a reasonable example: 
https://editor.swagger.io/

Sure there are some more examples online too to provide some guidance. We want to provide as much details as possible here as to what is returned to the client.

doc/swagger/swagger.yml
```diff
@@ -0,0 +1,792 @@
...
+  title: Acyl
+  description: The Acyl API provides authenticated users an array of options to create and manage dynamic environments.
+  contact:
+    name: Benjamen Keroack
+    email: benjamen@dollarshaveclub.com
+  license:
+    name: MIT
+    url: "https://opensource.org/licenses/MIT"
+  version: v0.7.6
+tags:
```
mikeykhalil, line 22: Can we add more details to the descriptions of each of the `v0`, `v1`, and `v2` tags? If possible, it could be helpful to have a general overview of the changes made for `v1` and `v2`.

doc/swagger/swagger.yml
wwalexander: I think this could be done something like

```
oneOf:
  - type: array
     items:
         type: string
  - $ref: '#/definitions/Envs'
    
```

which doesn't encode the condition for each type to be returned, but that can be added as part of the description if necessary.

doc/swagger/swagger.yml
wwalexander: The v0QAEnvironment has a [lot of fields and sub-fields](https://github.com/dollarshaveclub/acyl/blob/0ccbe5555a438d1630f0e906d9516c4c7671416e/pkg/api/api_v0.go#L37). I'm surprised there isn't a common way to automatically generate Swagger type definitions from Go structs...[go-swagger](https://github.com/go-swagger/go-swagger) exists but it has a larger scope.

wwalexander:
I think that the `pr` query parameter type could be changed to an `integer`, but maybe that's coupling it too tightly with Github choosing to identify PRs by integers. Also I think the `status` query parameter could potentially be an explicit enum:

```yaml
- name: status
  in: query
  description: enviornment status
  required: true
  type: string
  enum:
    - option1
    - option2
```

doc/swagger/swagger.yml
esterlinej: The goal of this isn't to host a server, but utilize this npm package to output yaml -> html, so there may be limitations as to what we can achieve.

doc/swagger/swagger.yml
esterlinej: It appears this package supports 2.0, and that's a 3.0 feature.

doc/swagger/swagger.yml
esterlinej: Added this as a definition, linking to the correct version for each version tag.

esterlinej:
Screen Shot 2020-09-28 at 6 02 19 PM
Screen Shot 2020-09-28 at 6 02 36 PM

doc/swagger/swagger.yml
```diff
@@ -0,0 +1,820 @@
...
+        maximum: 299
+  ErrorModel:
+    type: object
+    required:
+      - message
+      - code
+    properties:
+      message:
+        type: string
+  ClientError:
```
mikeykhalil, line 801: I don't think Acyl returns a JSON encoded object on most client errors. I believe it just returns an empty body, or a single string (e.g. `not found` or `forbidden`). So I don't think this definition is necessary and all references to it should likely be removed. Otherwise, clients and users might expect the response body to contain this information.

doc/swagger/swagger.yml
```diff
@@ -0,0 +1,820 @@
...
+    allOf:
+      - $ref: '#/definitions/ErrorModel'
+      - type: object
+        description: Represents a client error response
+        properties:
+          code:
+            type: integer
+            minimum: 400
+            maximum: 499
+  ServerError:
```
mikeykhalil, line 811: I don't think Acyl returns a JSON encoded object on most server errors. I believe it just returns an empty body. So I don't think this definition is necessary and all references to it should likely be removed.

doc/swagger/swagger.yml
```diff
@@ -0,0 +1,820 @@
...
+        required:
+          - array
+          - code
+  EventLog:
+    description: A representation of an event log response
+    type: object
+    required:
+      - object
+      - code
+    properties:
```
mikeykhalil, line 788: The EventLog doesn't have a `code` property. 

https://github.com/dollarshaveclub/acyl/blob/master/pkg/models/eventlogs.go#L17

doc/swagger/swagger.yml
```diff
@@ -0,0 +1,820 @@
...
+          SourceRef:
+            type: string
+  Env:
+    description: A representation of an individual environment response
+    type: object
+    required:
+      - object
+      - code
+    properties:
+      code:
```
mikeykhalil, line 770: Don't think and env has a `code` property either

esterlinej:
> I think that the `pr` query parameter type could be changed to an `integer`, but maybe that's coupling it too tightly with Github choosing to identify PRs by integers. Also I think the `status` query parameter could potentially be an explicit enum:
> 
> ```yaml
> - name: status
>   in: query
>   description: environment status
>   required: true
>   type: string
>   enum:
>     - UnknownStatus
>     - Spawned
>     - Success
>     - Failure
>     - Destroyed
>     - Updating
> ```

fixed, good catch

mikeykhalil (changes requested):
Found a few more places where object definitions contained http status codes. I called out a few instances that I saw, but would recommend reviewing to make sure there aren't any additional cases.

doc/swagger/swagger.yml
```diff
@@ -0,0 +1,820 @@
...
+          type: string
+        - name: repo
+          in: query
+          description: repository name
+          required: true
+          type: string
+        - name: status
+          in: query
+          description: enviornment status
+          required: true
```
bkeroackdsc, line 87: I don't think any of these parameters are strictly required. You just need at least one IIRC.

doc/swagger/swagger.yml
```diff
@@ -0,0 +1,820 @@
...
+          schema:
+            $ref: '#/definitions/ServerError'
+  /envs/{name}:
+    get:
+      tags:
+        - backward_compatible
+      summary: get enviornment details
+      operationId: getEnvsName
+      parameters:
+        - name: API-Key
```
bkeroackdsc, line 124: Is there a way to have a single "Authorization" section that describes API key authentication, instead of duplicating this parameter for every endpoint?

doc/swagger/swagger.yml
```diff
@@ -0,0 +1,820 @@
...
+        - name: name
+          in: path
+          description: environment name (e.g. `foo-bar-baz`)
+          required: true
+          type: string
+      responses:
+        200:
+          description: returns requested environment
+          schema:
+            $ref: '#/definitions/v0QAEnvironment'
```
bkeroackdsc, line 138: Instead of naming these directly after the Go types, I would make them a little more reader-friendly. Something like `QAEnvironment (v0)`.

doc/swagger/swagger.yml
```diff
@@ -0,0 +1,820 @@
...
+          in: header
+          description: API key or aborts with Unauthorized
+          required: true
+          type: string
+        - name: name
+          in: path
+          description: environment name (e.g. `foo-bar-baz`)
+          required: true
+          type: string
+      responses:
```
bkeroackdsc, line 165: This returns 404 if the environment isn't found I think.

doc/swagger/swagger.yml
```diff
@@ -0,0 +1,820 @@
...
+          description: no content
+        400:
+          description: bad request
+          schema:
+            $ref: '#/definitions/ClientError'
+        500:
+          description: internal server error
+          schema:
+            $ref: '#/definitions/ServerError'
+  /envs/{name}/success:
```
bkeroackdsc, line 176: Wow, these still exist! These endpoints (success, failure, event) are holdovers from the original Lambda-based DQA implementation.

I would mark these with big DEPRECATED warnings. I don't think they actually do anything (probably stubbed out).

doc/swagger/swagger.yml
```diff
@@ -0,0 +1,820 @@
...
+        - v0
+      summary: get list of environments, names only or full details, trailing slash `/v0/envs/` supported
+      operationId: getV0Envs
+      parameters:
+        - name: API-Key
+          in: header
+          description: API key or aborts with Unauthorized
+          required: true
+          type: string
+        - name: full_details
```
bkeroackdsc, line 255: I would mark `full_details` as deprecated. It basically results in a full table dump, which is very inefficient and slow (I believe several seconds of latency).

Clients should use the `_search` endpoint instead.

doc/swagger/swagger.yml
```diff
@@ -0,0 +1,820 @@
...
+      Message:
+        type: string
+  RefMap:
+    description: A representation of RefMap object
+    type: object
+    properties:
+      string:
+        type: string
+  AminoServiceToPort:
+    description: A representation of AminoServiceToPort object
```
bkeroackdsc, line 700: I would mark all Amino fields as deprecated and unused.

doc/swagger/swagger.yml
```diff
@@ -0,0 +1,820 @@
...
+  QAEnvironmentEvent:
+    description: A representation of QAEnvironmentEvent object
+    type: object
+    properties:
+      Timestamp:
+        type: string
+      Message:
+        type: string
+  RefMap:
+    description: A representation of RefMap object
```
bkeroackdsc, line 694: Explain what a RefMap is (a mapping of repo to git ref <branch>).

bkeroackdsc (changes requested):

doc/swagger/swagger.yml
esterlinej: Okay, I will comment that in the API files themselves as well as in the swagger docs

esterlinej:
Screen Shot 2020-09-29 at 1 10 10 PM
Screen Shot 2020-09-29 at 1 10 04 PM

doc/swagger/swagger.yml
```diff
@@ -0,0 +1,684 @@
...
+    required:
+      - object
+  Envs:
+    allOf:
+      - $ref: '#/definitions/Env'
+      - type: array
+        description: A representation of an array of environments response per the API version requested
+        required:
+          - array
+  EventLog:
```
mikeykhalil, line 666: Still think there is room to list the properties of an `EventLog` here. I'm not sure if we necessarily need to go into full detail here (since there is a lot of data that is more easily found in other endpoints), but at least mentioning the response contains a `log` property of type `[]string` would be helpful for users so they can parse the response accordingly.

mikeykhalil (changes requested):

doc/swagger/swagger.yml
esterlinej: Sure, will update.

doc/swagger/swagger.yml
```diff
@@ -0,0 +1,703 @@
...
+          required: false
+          type: string
+        - name: user
+          in: query
+          description: github user
+          required: false
+          type: string
+      responses:
+        200:
+          description: returns list of detailed environments as defined by `QAEnvironment (v2)`
```
mikeykhalil, line 517: The description says `returns list of detailed environments as defined by `QAEnvironment (v2)`,
but the schema for the same response code is `'#/definitions/Envs'`

Should the schema instead be an array of `'#/definitions/QAEnvironment_v2'? Or does the description need to be updated?

doc/swagger/swagger.yml
```diff
@@ -0,0 +1,703 @@
...
+          required: false
+          type: string
+        - name: user
+          in: query
+          description: github user
+          required: false
+          type: string
+      responses:
+        200:
+          description: returns list of detailed environments as defined by `QAEnvironment (v0)`
```
mikeykhalil, line 99: The description says returns list of detailed environments as defined by QAEnvironment (v0), but the schema for the same response code is `'#/definitions/Envs'`

Should the schema instead be an array of `'#/definitions/QAEnvironment_v0'`? Or does the description need to be updated?

doc/swagger/swagger.yml
```diff
@@ -0,0 +1,703 @@
...
+          schema:
+            $ref: '#/definitions/Envs'
```
mikeykhalil, line 273: The description says returns list of detailed environments as defined by QAEnvironment (v9), but the schema for the same response code is '#/definitions/Envs'`

Should the schema instead be an array of `'#/definitions/QAEnvironment_v0'? Or does the description need to be updated?

doc/swagger/swagger.yml
```diff
@@ -0,0 +1,703 @@
...
+        - name: user
+          in: query
+          description: github user
+          required: false
+          type: string
+      responses:
+        200:
+          description: returns list of detailed environments as defined by `QAEnvironment (v1)`
+          schema:
+            $ref: '#/definitions/Envs'
```
mikeykhalil, line 424: The description says returns list of detailed environments as defined by QAEnvironment (v1), but the schema for the same response code is '#/definitions/Envs'`

Should the schema instead be an array of `'#/definitions/QAEnvironment_v1'? Or does the description need to be updated?

doc/swagger/swagger.yml
```diff
@@ -0,0 +1,703 @@
...
+        - name: days
+          in: query
+          description: number of days (default `1`)
+          required: false
+          type: string
+      responses:
+        200:
+          description: returns list of detailed environments as defined by `QAEnvironment (v1)`
+          schema:
+            $ref: '#/definitions/Envs'
```
mikeykhalil, line 447: The description says returns list of detailed environments as defined by QAEnvironment (v1), but the schema for the same response code is '#/definitions/Envs'`

Should the schema instead be an array of `'#/definitions/QAEnvironment_v1'? Or does the description need to be updated?

mikeykhalil (changes requested):
Looks good to me!

doc/swagger/swagger.yml
esterlinej: We could update this, the Envs represent an array of these objects, it was meant to be more generic to cover all of the same types

esterlinej:
Screen Shot 2020-09-29 at 4 19 06 PM
Screen Shot 2020-09-29 at 4 15 43 PM

mikeykhalil (approved):

bkeroackdsc (approved):

Issue #65, opened 2020-01-21
Failure reports not emitted


bkeroackdsc:
At some point, [environment failure reports](https://github.com/dollarshaveclub/acyl/blob/master/pkg/nitro/env/env.go#L726-L757) stopped being emitted. We need to investigate and fix.

bkeroackdsc:
This could be a Metahelm issue: https://github.com/dollarshaveclub/metahelm/blob/master/pkg/metahelm/error.go

bkeroackdsc:
Example log:
```metahelm returned an error but it's not a ChartError (type: errors.operationError): not producing a failure report```

This was likely introduced when user/system errors were implemented, and may be related to #66

bkeroackdsc:
Failure reports have been refactored into a UI view, so this is no longer valid.

PR #121, opened 2020-09-29
fix: propagate chart error

mikeykhalil:
**Summary** 
This removes the feature of pushing detailed error reports to s3. This feature has been broken since at least January (https://github.com/dollarshaveclub/acyl/issues/65) and now that Acyl has a UI, generating some html that is served from S3 is no longer really necessary. 

Additionally, this PR fixes the underlying issue of preserving a ChartError, which is returned from the [metahelm library](https://github.com/dollarshaveclub/metahelm/blob/d4a4149401a2d17f184f264fb65b0861e09628a7/pkg/metahelm/error.go#L28). This was done by modifying the `operationError` type found in `"github.com/dollarshaveclub/acyl/pkg/nitro/errors"`. `operationErrors` now directly reference (and do not modify) the `metahelm.ChartError` and also implement the `Unwrap` method. These 2 changes allow us to reference the `ChartError`, from an `operationError` by doing the following: 
```
var ce metahelm.ChartError
if errors.As(err, &ce) { ... }
```

Fortunately, the `nitro/errors` package already had some tests setup and are still passing after the change of this implementation. 

Finally, this change persists the `ChartError` to Postgres. This will allow us to leverage Acyl's UI to display these errors, which can be helpful to users. 

---
One thing that I would like to point out that isn't obvious in this PR is that the [metahelm.ChartError type](https://github.com/dollarshaveclub/metahelm/blob/d4a4149401a2d17f184f264fb65b0861e09628a7/pkg/metahelm/error.go#L28) does not contain any json struct tags. So if it's preferred that the json fields are in snake case (to match the rest of the `EventStatusSummaryConfig` type), then we may want to update the metahelm library first and then update this PR accordingly.


pkg/nitro/env/env.go
```diff
@@ -1,17 +1,15 @@
 package env
 
 import (
-	"bytes"
 	"context"
+	stdliberrors "errors"
```
mikeykhalil, line 5: saw this naming convention being used when there were multiple "errors" packages being imported, but happy to update the package name of other imported packages if we prefer.

esterlinej (approved):
LGTM, that is a big AWS vendor package!

esterlinej (approved):
make it so

pkg/nitro/env/env.go
```diff
@@ -256,7 +246,11 @@ func (m *Manager) lockingOperation(ctx context.Context, repo string, pr uint, f
 		err = ctx.Err()
 	}
 	if err != nil {
-		eventlogger.GetLogger(ctx).SetCompletedStatus(models.FailedStatus)
+		var ce metahelmlib.ChartError
+		if stdliberrors.As(err, &ce) {
+			m.log(ctx, "error returned was a ChartError")
```
bkeroackdsc, line 251: You probably want the `SetFailedStatus()` method call within this conditional, right?

bkeroackdsc (approved):
LGTM. Love to see all that code deleted!

pkg/nitro/env/env.go
mikeykhalil: For the sake of simplicity, I intentionally placed `SetFailedStatus` outside of this conditional, since it should work in either case. I just made a call to `m.log` in the conditional for debugging purposes. 

If `stdliberrors.As(err, &ce)` returns true, `ce` will be populated with all the data from the original `ChartError` returned from the metahelm library. 
If `stdliberrors.As(err, &ce)` returns false, `ce` will just be the zero value, and the call to `SetFailedStatus` is essentially the same as `SetCompletedStatus(models.FailedStatus)`.

PR #122, opened 2020-10-01
Add Help View

esterlinej:
### Help View
- Implement a “Help” view which contains links to documentation and utilities that are helpful for developing/testing environment configs.
- Use one of these icons in the main navigation bar for the Help view (from Font Awesome):
- Link to Swagger API documentation
### Added Feature
- Support custom urls, `Added Resources` urls


esterlinej:
Screen Shot 2020-09-30 at 7 27 40 PM
Screen Shot 2020-09-30 at 7 27 08 PM

cmd/mockui.go
```diff
@@ -55,6 +55,7 @@ func addUIFlags(cmd *cobra.Command) {
 	cmd.PersistentFlags().StringVar(&serverConfig.UIPath, "ui-path", "/opt/ui", "Local filesystem path to UI assets")
 	cmd.PersistentFlags().StringVar(&serverConfig.UIBaseRoute, "ui-base-route", "/ui", "Base prefix for UI HTTP routes")
 	cmd.PersistentFlags().StringVar(&serverConfig.UIBrandingJSON, "ui-branding", string(brj), "Branding JSON configuration (see doc)")
+	cmd.PersistentFlags().StringSliceVar(&serverConfig.UIExternalDocumentURLs, "ui-doc-urls", []string{}, "External documentation URLs")
```
mikeykhalil, line 58: Do we need a value for this in the helm chart?

esterlinej: We can, it's not required at this time, but will update the helm chart after this before we deploy this release to add our confluence docs

esterlinej: 2nd thought, you're right, will add.

ui/assets/help.js
```diff
@@ -0,0 +1,59 @@
+function showAdditionalResourcesHeading() {
+    let arh = document.getElementById("hAdditionalResources");
+    arh.removeAttribute("class");
```
mikeykhalil, line 3: how come we are removing the class entirely here?

ui/assets/help.js
```diff
@@ -0,0 +1,59 @@
+function showAdditionalResourcesHeading() {
+    let arh = document.getElementById("hAdditionalResources");
```
mikeykhalil, line 2: Perhaps we should do a null check here as well?
```
if (arh == null) {
        return;
    }
```

ui/assets/help.js
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// getExternalDocumentURLs gets external document urls
+function getExternalDocumentURLs(env_name) {
+    let req = new XMLHttpRequest();
+
+    req.open('GET', `${apiBaseURL}/v2/help/document/urls`, true);
+    req.onload = function (e) {
+        if (req.status !== 200) {
+            console.log(`namespace pods request failed: ${req.status}: ${req.responseText}`);
```
mikeykhalil, line 38:
```suggestion
 console.log(`help document request failed: ${req.status}: ${req.responseTest}`);
```

ui/assets/help.js
```diff
@@ -0,0 +1,59 @@
...
+        let data = JSON.parse(req.response);
+        if (data.hasOwnProperty('document_urls')) {
+            if (Object.keys(data.document_urls).length > 0) {
+                showAdditionalResourcesHeading();
+                renderExternalDocumentURLs(data.document_urls);
+            }
+        }
+    };
+    req.onerror = function (e) {
+        console.error(`error getting namespace pod endpoint for ${env_name}: ${req.statusText}`);
```
mikeykhalil, line 51:
```suggestion
console.error(`error getting help documents: ${req.statusText}`);
```

pkg/api/api_v2.go
```diff
@@ -1083,3 +1086,24 @@ func (api *v2api) userEnvPodLogsHandler(w http.ResponseWriter, r *http.Request)
 		w.WriteHeader(http.StatusInternalServerError)
 	}
 }
+
+type V2ExternalDocumentURLs struct {
+	DocumentURLs []string `json:"document_urls"`
+}
+
+// externalDocumentURLsHandler returns list of external document urls from server configuration
+func (api *v2api) externalDocumentURLsHandler(w http.ResponseWriter, r *http.Request) {
```
mikeykhalil, line 1095: Since these values don't change during the lifetime of the server, could we just render the urls when we go to serve the html/javascript assets? Or are there other benefits for making an endpoints for this?

ui/assets/help.js
esterlinej: The class by default bootstap is hidden, so to show it, we need to remove the class

pkg/api/api_v2.go
esterlinej: Slack Huddle...

esterlinej: Updated, thanks for the brain storm, resolving

pkg/api/api_ui.go
```diff
@@ -109,7 +110,7 @@ func newSessionsCookieStore(oauthCfg OAuthConfig) sessions.Store {
 	return cstore
 }
 
-func newUIAPI(baseURL, assetsPath, routePrefix string, reload bool, branding config.UIBrandingConfig, dl persistence.DataLayer, oauthCfg OAuthConfig, logger *log.Logger) (*uiapi, error) {
+func newUIAPI(baseURL, assetsPath, routePrefix string, reload bool, branding config.UIBrandingConfig, dl persistence.DataLayer, oauthCfg OAuthConfig, logger *log.Logger, docUrls []string) (*uiapi, error) {
```
bkeroackdsc, line 113: `docURLs` can probably be put into UIBrandingConfig.

pkg/config/config.go
```diff
@@ -33,6 +33,7 @@ type ServerConfig struct {
 	UIPath                     string
 	UIBaseRoute                string
 	UIBrandingJSON             string
+	UIExternalDocumentURLs     []string
```
bkeroackdsc, line 36: This should be a map of name to URL, which then gets templated like this: 
`<a href="{{ url }}">{{ name }}</a>`

ui/views/help.html
```diff
@@ -0,0 +1,50 @@
...
+                        </ul>
+                        {{ end }}
+                    </strong>
+                </div>
+            </div>
+        </div>
+        {{ template "footer-scripts" . }}
+        <script
+                type="text/javascript"
+                src="{{ .APIBaseURL }}/ui/static/help.js"
```
bkeroackdsc, line 47: There is no help.js so you can get rid of this.

cmd/mockui.go
bkeroackdsc, line 58: Since we need both link name and URL (not just URL as you have here), I would either do this as a string slice of `<name>=<url>` strings, or as a single JSON-encoded array of objects: `[{"name": "foo", "url": "bar"}]`.

You could even extend the JSON branding configuration (the option above) to include additional documentation URLs.

bkeroackdsc (changes requested):

.helm/charts/acyl/values.yaml
```diff
@@ -44,6 +44,7 @@ app:
       favicon_url: ''
       title: ''
       title_style: ''
+    document_urls: ''
```
bkeroackdsc: Incorrect indentation

cmd/mockui.go
```diff
@@ -168,9 +168,11 @@ func mockui(cmd *cobra.Command, args []string) {
 	serverConfig.UIBaseURL = "http://" + listenAddr
 
 	var branding config.UIBrandingConfig
+	log.Println(serverConfig.UIBrandingJSON)
```
bkeroackdsc, line 171: Remove debug printlns

pkg/config/config.go
```diff
@@ -204,15 +204,16 @@ type SecretsConfig struct {
...
+	FaviconURL        string            `json:"favicon_url"`
+	LogoURL           string            `json:"logo_url"`
+	LogoHeight        string            `json:"logo_height"`
+	LogoWidth         string            `json:"logo_width"`
+	LogoStyle         string            `json:"logo_style"`
+	LogoLink          string            `json:"logo_link"`
+	LogoAlt           string            `json:"logo_alt"`
+	Title             string            `json:"title"`
+	TitleStyle        string            `json:"title_style"`
+	AdditionalDocUrls map[string]string `json:"additional_doc_urls"`
```
bkeroackdsc, line 216: Naming: `AdditionalDocURLs`

bkeroackdsc (changes requested):

bkeroackdsc (approved):

PR #124, opened 2020-10-01
Use array of event IDs to retrieve environment events

wwalexander:
Instead of retrieving event logs by looking up event logs with a given environment name, this PR adds a column event_ids to the qa_environments table that stores the event foreign keys to allow for fast looks. Currently, this PR backfills previous env_name-based data into the new column, and maintains the env_name column in the event_logs table alongside the event_ids array.

### DIFF
--- a/pkg/persistence/datalayer_test.go
+++ b/pkg/persistence/datalayer_test.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"log"
 	"net"
 	"strings"
 	"testing"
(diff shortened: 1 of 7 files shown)


pkg/persistence/datalayer_test.go
```diff
@@ -901,9 +902,14 @@ func TestDataLayerGetMostRecentOrdering(t *testing.T) {
 		t.Fatalf("should have returned all envs: %v", qas)
 	}
 	now := time.Now().UTC()
+	for _, qa := range qas {
+		log.Println(qa.Created.String())
+	}
```
bkeroackdsc, line 907: Looks like some debugging printlns.

bkeroackdsc (approved):
LGTM

pkg/persistence/datalayer_test.go
wwalexander: Fixed by 72819f6.

PR #126, opened 2020-10-02
Help Page Enhancements

esterlinej:
- add breadcrums
- google no translate


esterlinej:
Screen Shot 2020-10-05 at 11 27 12 AM

mikeykhalil:
Do we currently make use of the `lang` attribute? This should inform browsers about the language used in the page. 
https://www.w3schools.com/tags/ref_language_codes.asp

esterlinej:
> Do we currently make use of the `lang` attribute? This should inform browsers about the language used in the page.
> https://www.w3schools.com/tags/ref_language_codes.asp

Will test this out in DQAs, if works will utilize this instead.

esterlinej:
Screen Shot 2020-10-05 at 11 53 32 AM

ui/views/help.html
```diff
@@ -1,43 +1,62 @@
-<html>
+<html lang="en">
     <head>
         {{ template "header-head" . }}
         <script type="text/javascript">
             const apiBaseURL = "{{ .APIBaseURL }}";
         </script>
+        <style>
```
bkeroackdsc, line 7: I'm fine with this but if it gets any more complex (this goes for other views too), we should split out to a separate CSS file (and then see if we can de-dupe into a set of shared styles in a partial).

bkeroackdsc (approved):

mikeykhalil (approved):

PR #127, opened 2020-10-02
Use event_ids array in GetEventLogsWithStatusByEnvName

wwalexander:
I updated GetEventLogsByEnvName to use the new schema, but didn't see GetEventLogsWithStatusByEnvName, which just adds an extra column to the query and is what is actually called by the API. This PR makes the two methods use the same query structure.


cleanshavenalex (approved):
lgtm

bkeroackdsc (approved):

wwalexander:
I've added a duplicate check when adding event logs to an environment after realizing a temporary issue where I was inserting a duplicate event. The issue is fixed now, but I figure the check is worth keeping the DB in a sane state if an event is for some reason repeatedly associated with an environment.

pkg/persistence/datalayer_test.go
```diff
@@ -1213,6 +1214,11 @@ func TestDataLayerGetEventLogsByEnvName(t *testing.T) {
 		t.Fatalf("should have succeeded: %v", err)
 	}
 	if i := len(logs); i != 2 {
+		for _, l := range logs {
+			log.Println(l.ID)
+			log.Println(l.Created)
+		}
+		t.Errorf("%v", logs) // DELETEME
```
bkeroackdsc, line 1221: `DELETEME`

bkeroackdsc (approved):

pkg/persistence/datalayer_test.go
wwalexander: Ah, apologies! Fixed by 6c4a72d.

PR #128, opened 2020-10-06
Acyl Feature: API Keys

esterlinej:
### API Keys
Currently Acyl has a list of API keys (UUIDs) that are read from Vault at startup and then checked via the auth request middleware in pkgs/api/middleware.go.

In preparation for user-managed API keys, we need to create a new table in the database (api_keys) with the following columns (_add anything that's missing from this list as needed_):
- API key (random UUID), this should be the primary key
- created & last used timestamps
- permissions level (probably an integer-based enum: read-only, read-write, or admin)
- name string
- description string
- github user string

Models and Datalayer methods + tests will need to be implemented for basic CRUD operations: 
- Create
- Get by ID
- Get by GitHub User (list of API keys)
- Delete
- Update last used timestamp (everything else should be immutable).

Models and Datalayer methods + tests will need to be implemented for basic CRUD operations: Create, Get, Delete, Update last used timestamp only (everything else should be immutable).

**We should continue to support the legacy list of static API keys from Vault**, but in addition to this, every time an API call is made we should also check the API key supplied by the client against the keys in the database, and authorize the API call if a database API key matches and has sufficient permissions.

For now, all endpoints except `DELETE /envs/<name>` should be considered read-only and require read-only key permissions. The delete environment endpoint should be considered read-write. There are no admin-only endpoints currently. Static API keys from Vault should be treated as admin keys.

### tokenAuth
- refactored `authRequest` into `tokenAuth` combining the master keys check, then processing user generated keys.


pkg/models/apikeys.go
```diff
@@ -0,0 +1,74 @@
...
+
+func (apik APIKey) InsertParams() string {
+	params := []string{}
+	for i := range strings.Split(apik.InsertColumns(), ",") {
+		params = append(params, fmt.Sprintf("$%v", i+1))
+	}
+	return strings.Join(params, ", ")
+}
+
+type PermissionsLevels int
```
bkeroackdsc, line 46: Name should be `PermissionLevel` (singular).

pkg/models/apikeys.go
```diff
@@ -0,0 +1,74 @@
...
+
+const (
+	ReadOnly PermissionsLevels = iota
+	Write
+	Admin
+)
```
bkeroackdsc, line 52: Enums should usually start with an unknown value so that the zero value can be distinguished.

The names should also include the type because these are package-level types and thus "ReadOnly" is not descriptive enough within the entire `models` package.

```go
const (
   UnknownPermission PermissionLevel = iota
   ReadOnlyPermission
   WritePermission
   AdminPermission
)
```

pkg/models/apikeys.go
```diff
@@ -0,0 +1,74 @@
...
+
+func (apik APIKey) DecodePermissionsLevel(pl PermissionsLevels) string {
```
bkeroackdsc, line 54: Use `go generate` with stringer here to autogenerate strings from enums.

pkg/persistence/apikeys.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/pkg/errors"
+
+	"github.com/dollarshaveclub/acyl/pkg/models"
+)
+
+// CreateAPIKey creates a new api key for a user from the APIKey model
+func (pg *PGLayer) CreateAPIKey(ctx context.Context, permissionsLevel, name, description, githubUser string) (uuid.UUID, error) {
+	var id uuid.UUID
+	in := &models.APIKey{
+		APIKey: uuid.New(),
```
bkeroackdsc, line 19: This should be `uuid.NewRandom()` with the error checked.

pkg/persistence/apikeys.go
```diff
@@ -0,0 +1,87 @@
...
+		Created: time.Now().UTC().Truncate(1 * time.Microsecond),
+		LastUsed: pq.NullTime{Time: time.Now().UTC().Truncate(1 * time.Microsecond), Valid: true},
```
bkeroackdsc, line 21: I don't think you need to truncate times here.

pkg/persistence/apikeys.go
```diff
@@ -0,0 +1,87 @@
...
+		Name: name,
+		Description: description,
+		GitHubUser: githubUser,
+	}
+	in.PermissionsLevel = in.EncodePermissionsLevel(permissionsLevel)
```
bkeroackdsc, line 26: You're converting to string but the column type is integer.

pkg/persistence/apikeys.go
```diff
@@ -0,0 +1,87 @@
...
+	q := `INSERT INTO api_keys (` + in.InsertColumns() + `) VALUES (` + in.InsertParams() + `) RETURNING id;`
+	if err := pg.db.QueryRowContext(ctx, q, in.InsertValues()...).Scan(&id); err != nil {
```
bkeroackdsc, line 28: This can be an `Exec()` since you already have the id so the "RETURNING ..." clause is unnecessary.

pkg/persistence/apikeys.go
```diff
@@ -0,0 +1,87 @@
...
+		}
+		out = append(out, &key)
+	}
+	return out, nil
+}
+
+// UpdateAPIKeyLastUsed updates the last used field for the api key id
+func (pg *PGLayer) UpdateAPIKeyLastUsed(ctx context.Context, id uuid.UUID) error {
+	q := `UPDATE api_keys SET last_used = $1 WHERE id = $2;`
+	_, err := pg.db.ExecContext(ctx, q, pq.NullTime{Time: time.Now().UTC().Truncate(1 * time.Microsecond), Valid: true}, id)
```
bkeroackdsc, line 72: You don't need to truncate time.

pkg/persistence/fake.go
```diff
@@ -27,6 +28,7 @@ type lockingDataMap struct {
 	k8s        map[string]*models.KubernetesEnvironment
 	elogs      map[uuid.UUID]*models.EventLog
 	uisessions map[int]*models.UISession
+	apikeys    map[string][]*models.APIKey
```
bkeroackdsc, line 31: The map key should be `uuid.UUID`.

pkg/persistence/fake.go
```diff
@@ -1169,3 +1172,76 @@ func (fdl *FakeDataLayer) DeleteExpiredUISessions() (uint, error) {
 	}
 	return uint(len(rmkeys)), nil
 }
+
+func (fdl *FakeDataLayer) CreateAPIKey(ctx context.Context, permissionsLevel, name, description, githubUser string) (uuid.UUID, error) {
+	fdl.doDelay()
+	fdl.data.Lock()
+	defer fdl.data.Unlock()
+	key := &models.APIKey{
+		APIKey: uuid.New(),
```
bkeroackdsc, line 1181: `uuid.NewRandom()`

pkg/persistence/fake.go
```diff
@@ -1169,3 +1172,76 @@ func (fdl *FakeDataLayer) DeleteExpiredUISessions() (uint, error) {
...
+		Created: time.Now().UTC().Truncate(1 * time.Microsecond),
```
bkeroackdsc, line 1182: Don't truncate time.

pkg/persistence/fake.go
```diff
@@ -1169,3 +1172,76 @@ func (fdl *FakeDataLayer) DeleteExpiredUISessions() (uint, error) {
...
+func (fdl *FakeDataLayer) GetAPIKeysByGithubUser(ctx context.Context, githubUser string) ([]*models.APIKey, error) {
+	fdl.doDelay()
+	fdl.data.Lock()
+	defer fdl.data.Unlock()
+	keys := fdl.data.apikeys[githubUser]
+	if len(keys) == 0 {
+		return nil, errors.Errorf("error getting api keys for user: %v", githubUser)
+	}
+	return keys, nil
+}
```
bkeroackdsc, line 1219: You have this backward. Map key should be UUID, and to find keys by user you should iterate through the data set.

pkg/persistence/fake.go
```diff
@@ -1169,3 +1172,76 @@ func (fdl *FakeDataLayer) DeleteExpiredUISessions() (uint, error) {
...
+
+func (fdl *FakeDataLayer) UpdateAPIKeyLastUsed(ctx context.Context, apiKey uuid.UUID) error {
+	fdl.doDelay()
+	fdl.data.Lock()
+	defer fdl.data.Unlock()
+	for _, keys := range fdl.data.apikeys {
+		for _, v := range keys {
+			if v.APIKey == apiKey {
+				v.LastUsed = pq.NullTime{Time: time.Now().UTC().Truncate(1 * time.Millisecond), Valid: true}
```
bkeroackdsc, line 1228: Truncating time again.

pkg/persistence/fake.go
```diff
@@ -1169,3 +1172,76 @@ func (fdl *FakeDataLayer) DeleteExpiredUISessions() (uint, error) {
...
+}
+
+func (fdl *FakeDataLayer) DeleteAPIKey(ctx context.Context, apiKey uuid.UUID) error {
+	fdl.doDelay()
+	fdl.data.Lock()
+	defer fdl.data.Unlock()
+	for _, keys := range fdl.data.apikeys {
+		for k, v := range keys {
+			if v.APIKey == apiKey {
+				fdl.data.apikeys[v.GitHubUser] = append(fdl.data.apikeys[v.GitHubUser][:k], fdl.data.apikeys[v.GitHubUser][k+1:]...)
```
bkeroackdsc, line 1242: This would just be a simple map key delete if the key was a UUID and not a github user.

bkeroackdsc (changes requested):
You're missing the functionality from the second half of the ticket. Specifically, authorizing API calls against keys that exist in the database (utilizing the datalayer methods you've implemented).

pkg/persistence/apikeys.go
esterlinej: This encode pl takes the string and converts it into an integer

pkg/api/api_v0.go
```diff
@@ -116,9 +116,9 @@ func (api *v0api) register(r *muxtrace.Router) error {
 	// backwards-compatible routes
 	r.HandleFunc("/spawn", middlewareChain(api.legacyGithubWebhookHandler, waitMiddleware.waitOnRequest)).Methods("POST")
 	r.HandleFunc("/webhook", middlewareChain(api.legacyGithubWebhookHandler, waitMiddleware.waitOnRequest)).Methods("POST")
-	r.HandleFunc("/envs/_search", middlewareChain(api.envSearchHandler, authMiddleware.authRequest)).Methods("GET")
-	r.HandleFunc("/envs/{name}", middlewareChain(api.envDetailHandler, authMiddleware.authRequest)).Methods("GET")
-	r.HandleFunc("/envs/{name}", middlewareChain(api.envDestroyHandler, authMiddleware.authRequest, waitMiddleware.waitOnRequest)).Methods("DELETE")
+	r.HandleFunc("/envs/_search", middlewareChain(authMiddleware.tokenAuth(api.envSearchHandler, models.ReadOnlyPermission))).Methods("GET")
+	r.HandleFunc("/envs/{name}", middlewareChain(authMiddleware.tokenAuth(api.envDetailHandler, models.ReadOnlyPermission))).Methods("GET")
+	r.HandleFunc("/envs/{name}", middlewareChain(authMiddleware.tokenAuth(api.envDestroyHandler, models.WritePermission), waitMiddleware.waitOnRequest)).Methods("DELETE")
```
bkeroackdsc, line 121: On second thought, this should require Admin permissions since otherwise it would allow any user to delete any environment (not just their own).

pkg/api/middleware.go
```diff
@@ -36,20 +37,34 @@ var authMiddleware = reqAuthorizor{}
...
+		key := r.Header.Get(apiKeyHeader)
+		if key != "" {
 			for _, k := range ra.apiKeys {
-				if h == k {
+				if key == k {
 					f(w, r)
 					return
 				}
 			}
+			apikey, err := authMiddleware.DL.GetAPIKeyById(r.Context(), uuid.MustParse(key))
```
bkeroackdsc, line 53: `uuid.MustParse()` will panic if the UUID can't be parsed (which would kill the whole process). Instead, parse and check the error.

pkg/persistence/apikeys.go
```diff
@@ -0,0 +1,90 @@
...
+
+// CreateAPIKey creates a new api key for a user from the ID model
+func (pg *PGLayer) CreateAPIKey(ctx context.Context, permissionLevel, name, description, githubUser string) (uuid.UUID, error) {
+	id, err := uuid.NewRandom()
+	if err != nil {
+		return uuid.Nil, errors.Wrap(err, "error creating new random uuid")
+	}
+	in := &models.APIKey{
+		ID:          id,
+		Created:     time.Now().UTC().Truncate(1 * time.Microsecond),
```
bkeroackdsc, line 23: You don't need to truncate time.

pkg/api/middleware.go
```diff
@@ -36,20 +37,34 @@ var authMiddleware = reqAuthorizor{}
...
+			if err != nil {
+				w.WriteHeader(http.StatusInternalServerError)
+				return
+			}
+			if apikey == nil {
+				w.WriteHeader(http.StatusNotFound)
+				return
+			}
+			if apikey.PermissionLevel >= minPermission {
+				f(w, r)
```
bkeroackdsc, line 63: Update the last used timestamp here.

pkg/api/middleware.go
```diff
@@ -36,20 +37,34 @@ var authMiddleware = reqAuthorizor{}
...
 			for _, k := range ra.apiKeys {
-				if h == k {
+				if key == k {
 					f(w, r)
 					return
 				}
 			}
+			apikey, err := authMiddleware.DL.GetAPIKeyById(r.Context(), uuid.MustParse(key))
+			if err != nil {
+				w.WriteHeader(http.StatusInternalServerError)
```
bkeroackdsc, line 55: Log the error.

pkg/api/middleware.go
```diff
@@ -36,20 +37,34 @@ var authMiddleware = reqAuthorizor{}
...
+				return
+			}
+			if apikey == nil {
+				w.WriteHeader(http.StatusNotFound)
```
bkeroackdsc, line 59: `StatusForbidden` not a 404.

pkg/persistence/datalayer.go
```diff
@@ -98,3 +99,11 @@ type UISessionsDataLayer interface {
 	GetUISession(id int) (*models.UISession, error)
 	DeleteExpiredUISessions() (uint, error)
 }
+
+type APIKeyDataLayer interface {
+	CreateAPIKey(ctx context.Context, permissionLevel, name, description, githubUser string) (uuid.UUID, error)
```
bkeroackdsc, line 104: `permissionLevel` should be `models.PermissionLevel` instead of string.

pkg/persistence/fake.go
```diff
@@ -1169,3 +1172,79 @@ func (fdl *FakeDataLayer) DeleteExpiredUISessions() (uint, error) {
...
+		return nil, nil
+	}
+	return keys, nil
+}
+
+func (fdl *FakeDataLayer) UpdateAPIKeyLastUsed(ctx context.Context, id uuid.UUID) error {
+	fdl.doDelay()
+	fdl.data.Lock()
+	defer fdl.data.Unlock()
+	for _, v := range fdl.data.apikeys {
```
bkeroackdsc, line 1232: Do a map lookup instead of iterating through it.

pkg/persistence/fake.go
```diff
@@ -1169,3 +1172,79 @@ func (fdl *FakeDataLayer) DeleteExpiredUISessions() (uint, error) {
...
+	fdl.data.apikeys[id] = key
+	return id, nil
+}
+
+func (fdl *FakeDataLayer) GetAPIKeyById(ctx context.Context, id uuid.UUID) (*models.APIKey, error) {
+	fdl.doDelay()
+	fdl.data.Lock()
+	defer fdl.data.Unlock()
+	out := &models.APIKey{}
+	for _, v := range fdl.data.apikeys {
```
bkeroackdsc, line 1201: Map lookup not iteration.

pkg/persistence/fake.go
```diff
@@ -1169,3 +1172,79 @@ func (fdl *FakeDataLayer) DeleteExpiredUISessions() (uint, error) {
...
+		}
+	}
+	return nil
+}
+
+func (fdl *FakeDataLayer) DeleteAPIKey(ctx context.Context, id uuid.UUID) error {
+	fdl.doDelay()
+	fdl.data.Lock()
+	defer fdl.data.Unlock()
+	for _, v := range fdl.data.apikeys {
```
bkeroackdsc, line 1244: Map lookup not iteration

bkeroackdsc (changes requested):

pkg/persistence/apikeys.go
esterlinej: Missed this one, will remove

pkg/api/middleware.go
```diff
@@ -36,14 +37,42 @@ var authMiddleware = reqAuthorizor{}
...
+				w.WriteHeader(http.StatusInternalServerError)
+				return
+			}
+			apikey, err := authMiddleware.DL.GetAPIKeyById(r.Context(), keyID)
+			if err != nil {
+				log("error getting api key: %v", err)
+				w.WriteHeader(http.StatusInternalServerError)
+				return
+			}
+			if apikey != nil {
```
esterlinej, line 68: Rather than return `Forbidden` on nil, I flipped this logic `if not nil` then check and update, otherwise it kicks down to `Unauthorized`, which was the default experience with `authRequest`

pkg/api/middleware.go
```diff
@@ -50,16 +53,29 @@ func (ra reqAuthorizor) tokenAuth(f http.HandlerFunc, minPermission models.Permi
 					return
 				}
 			}
-			apikey, err := authMiddleware.DL.GetAPIKeyById(r.Context(), uuid.MustParse(key))
+			keyID, err := uuid.Parse(key)
 			if err != nil {
+				log("error parsing api key id: %v", err)
+				w.WriteHeader(http.StatusInternalServerError)
+				return
```
bkeroackdsc: Bad/invalid API key should be a StatusForbidden error, not a 500.

pkg/api/middleware.go
```diff
@@ -36,14 +37,42 @@ var authMiddleware = reqAuthorizor{}
...
-				if h == k {
+				if key == k {
+					f(w, r)
+					return
+				}
+			}
+			keyID, err := uuid.Parse(key)
+			if err != nil {
+				log("error parsing api key id: %v", err)
+				w.WriteHeader(http.StatusInternalServerError)
```
bkeroackdsc, line 59: Error parsing the key (meaning bad/invalid key) should be a StatusForbidden, not a 500.

pkg/models/apikeys.go
```diff
@@ -0,0 +1,81 @@
...
+	case "read-only":
+		return ReadOnlyPermission
+	case "write":
+		return WritePermission
+	case "admin":
+		return AdminPermission
+	default:
+		return UnknownPermission
+	}
+}
```
bkeroackdsc, line 81: You don't need this encoding/decoding. Just use the enum values.

pkg/persistence/fake.go
```diff
@@ -1169,3 +1172,77 @@ func (fdl *FakeDataLayer) DeleteExpiredUISessions() (uint, error) {
...
+		Description:     description,
+		GitHubUser:      githubUser,
+	}
+	fdl.data.apikeys[id] = key
+	return id, nil
+}
+
+func (fdl *FakeDataLayer) GetAPIKeyById(ctx context.Context, id uuid.UUID) (*models.APIKey, error) {
+	fdl.doDelay()
+	fdl.data.Lock()
```
bkeroackdsc, line 1199: Since this is a read-only method, this should be `fdl.data.RLock()` and `RUnlock()`

pkg/persistence/fake.go
```diff
@@ -1169,3 +1172,77 @@ func (fdl *FakeDataLayer) DeleteExpiredUISessions() (uint, error) {
...
+	if !ok {
+		return nil, nil
+	}
+	out = *ak
+	return &out, nil
+}
+
+func (fdl *FakeDataLayer) GetAPIKeysByGithubUser(ctx context.Context, githubUser string) ([]*models.APIKey, error) {
+	fdl.doDelay()
+	fdl.data.Lock()
```
bkeroackdsc, line 1212: `RLock()` and `RUnlock()`

bkeroackdsc (changes requested):

bkeroackdsc (approved):

PR #131, opened 2020-10-07
fix: adding in ingress and values for multiple hosts

jmedberydsc:
- Allowing for multiple values to be passed into annotations
- Allowing for multiple values to be passed for hosts
- default settings for values

### DIFF
--- a/.helm/charts/acyl/templates/ingress.yaml
+++ b/.helm/charts/acyl/templates/ingress.yaml
@@ -17,7 +17,7 @@ spec:
     {{- range $host := .Values.ingress.hosts }}
     - host: {{ $host }}
       http:
-      	paths:
+        paths:
           - path: /
             backend:
               serviceName: {{ $serviceName }}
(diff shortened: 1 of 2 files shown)


.helm/charts/acyl/templates/ingress.yaml
```diff
@@ -7,18 +8,19 @@ metadata:
...
-    - host: {{ .Values.env_name }}-{{ .Values.ingress.argo_host }}
-    {{- else }}
-    - host: {{ .Values.ingress.argo_host }}
-    {{- end }}
+    {{- $serviceName := .Values.service.name -}}
+    {{- range $host := .Values.ingress.hosts }}
+    - host: {{ $host }}
       http:
-        paths:
+      	paths:
```
cleanshavenalex, line 20: is this too indented?

cleanshavenalex (commented):
left a comment

cleanshavenalex (approved):
+1

.helm/charts/acyl/templates/ingress.yaml
```diff
@@ -1,3 +1,4 @@
+{{- $servicePort := .Values.service.internalPort -}}
```
mikeykhalil, line 1: `servicePort` should be defined right below `serviceName`. This reduces the the scope of the variable and also reduces the distance between where it's defined and where it is used.

mikeykhalil (approved):
Left one small comment, but looks good to me!

mikeykhalil (approved):

PR #133, opened 2020-10-09
Add cancelled status when preemption is detected

wwalexander:
This commit additionally moves from github.com/pkg/errors to Go 1.13's native error-wrapping in pkg/nitro/env.


pkg/nitro/env/env.go
```diff
@@ -244,15 +245,22 @@ func (m *Manager) lockingOperation(ctx context.Context, repo string, pr uint, f
 		err = opErr
 	case <-ctx.Done():
 		err = ctx.Err()
+		cancelled = true
```
mikeykhalil, line 248: I wonder if we could just do `nitroerrors.CancelledError(ctx.Err())` here, as opposed to keeping track of this boolean.

Then, once we verify that `err != nil` below, we can do something like 

```
if nitroerrors.IsCanceledError(err) { 
  eventlogger.GetLogger(ctx).SetCompletedStatus(models.CancelledStatus)
} else { 
  eventlogger.GetLogger(ctx).SetFailedStatus(ce)
}
```

pkg/models/eventlogs.go
```diff
@@ -91,6 +91,7 @@ const (
 	PendingStatus
 	DoneStatus
 	FailedStatus
+	CancelledStatus
```
mikeykhalil, line 94: A comment earlier on in this file indicates that there is code that needs to be regenerated after this change
```
//go:generate stringer -type=EventStatus
```

wwalexander: Should be fixed by 9fe8e4c.

pkg/nitro/env/env.go
wwalexander: That makes more sense! Should be fixed by 9fe8e4c.

pkg/nitro/env/env.go
```diff
@@ -243,9 +242,11 @@ func (m *Manager) lockingOperation(ctx context.Context, repo string, pr uint, f
 	case opErr := <-ch:
 		err = opErr
 	case <-ctx.Done():
-		err = ctx.Err()
+		err = nitroerrors.CancelledError(ctx.Err())
 	}
-	if err != nil {
+	if nitroerrors.IsCancelledError(err) {
```
bkeroackdsc, line 247: You're wrapping the error, but then checking just a few lines down. Something like this would be simpler: https://gist.github.com/bkeroackdsc/f58b017235861eaa23461528939d89c8

pkg/nitro/env/env.go
```diff
@@ -243,9 +242,11 @@ func (m *Manager) lockingOperation(ctx context.Context, repo string, pr uint, f
...
+		eventlogger.GetLogger(ctx).SetCompletedStatus(models.CancelledStatus)
+	} else {
```
bkeroackdsc, line 249: I would do an early return here instead of the `else` clause.

ui/assets/env.js
```diff
@@ -109,6 +109,10 @@ function eventrow(event) {
             tr.className = "table-warning";
             tdstatus.innerHTML = `<span class="badge badge-warning">Pending</span>`;
             break;
+        case "cancelled":
+            tr.className = "table-dark";
+            tdstatus.innerHTML = `<span class="badge badge-dark">Cancelled</span>`;
+            break;
```
bkeroackdsc, line 115: We're not changing the Environment Status, just the event status.

bkeroackdsc (changes requested):

ui/assets/env.js
wwalexander: I think I'm missing something - this switch is on event.status so the cancelled status should have it's own representation, no?

pkg/nitro/env/env.go
wwalexander: Should be fixed by f668249.

pkg/nitro/env/env.go
wwalexander: Should be fixed by f668249.

ui/assets/env.js
bkeroackdsc: You're correct, my mistake.

bkeroackdsc (approved):
LGTM

PR #135, opened 2020-10-21
Use explicit errors to detect system/user/cancellation errors

wwalexander:


mikeykhalil (changes requested):
LGTM, just notice that a lot of the `if err != nil { ` blocks contain an extra trailing `\n`. So those should probably be removed to keep things consistent. 

I do remember @bkeroackdsc had some opinions and thoughts around classifying system errors versus user errors, so I would ensure to make sure he reviews before merging.

pkg/nitro/env/env.go
```diff
@@ -539,13 +554,15 @@ func (m *Manager) create(ctx context.Context, rd *models.RepoRevisionData) (envn
...
-		return "", errors.Wrap(err, "error enforcing global limit")
+		return "", fmt.Errorf("error enforcing global limit: %w", err)
+
 	}
 
 	chartSpan, ctx := tracer.StartSpanFromContext(ctx, "build_and_install_charts")
 	if err = m.CI.BuildAndInstallCharts(ctx, &metahelm.EnvInfo{Env: newenv.env, RC: newenv.rc}, mcloc); err != nil {
 		chartSpan.Finish(tracer.WithError(err))
-		return "", errors.Wrap(nitroerrors.UserError(err), "error installing charts")
+		return "", fmt.Errorf("error installing charts: %w", nitroerrors.UserError(err))
```
bkeroackdsc, line 564: This is another instance where you can't reliably tell at this point in the callstack whether this is a user or system error (might be a misconfigured helm chart, or the K8s control plane could be having problems, etc).

pkg/nitro/env/env.go
```diff
@@ -591,7 +608,7 @@ func (m *Manager) delete(ctx context.Context, rd *models.RepoRevisionData, reaso
 			m.log(ctx, "no extant envs for destroy request")
 			envs, err := m.DL.GetQAEnvironmentsByRepoAndPR(ctx, rd.Repo, rd.PullRequest)
 			if err != nil {
-				return errors.Wrapf(nitroerrors.SystemError(err), "error getting environments associated with the repo (%v) and PR (%v)", rd.Repo, rd.PullRequest)
+				return fmt.Errorf("error getting environments associated with the repo (%v) and PR (%v): %w", rd.Repo, rd.PullRequest, nitroerrors.SystemError(err))
```
bkeroackdsc, line 611: System error wrapping should rarely (if ever) be used. *Everything* is a system error, unless it's specifically annotated as a User Error.

pkg/nitro/env/env.go
```diff
@@ -603,7 +620,8 @@ func (m *Manager) delete(ctx context.Context, rd *models.RepoRevisionData, reaso
 			}
 			return nil
 		}
-		return errors.Wrap(nitroerrors.SystemError(err), "error getting extant environment")
+		return fmt.Errorf("error getting extant environment: %w", nitroerrors.SystemError(err))
```
bkeroackdsc, line 623: Same here. Eliminate all these SystemError() annotations. Perhaps just remove that annotation function entirely.

pkg/nitro/env/env.go
```diff
@@ -773,15 +792,16 @@ func (m *Manager) update(ctx context.Context, rd *models.RepoRevisionData) (envn
...
+			return "", fmt.Errorf("error getting helm releases for env: %w", nitroerrors.SystemError(err))
 		}
 		rsls := map[string]string{}
 		for _, r := range releases {
 			rsls[r.Name] = r.Release // chart title (dependency name) to release name
 		}
 		envinfo.Releases = rsls
 		if err := m.CI.BuildAndUpgradeCharts(ctx, envinfo, k8senv, mcloc); err != nil {
-			return envinfo.Env.Name, errors.Wrap(nitroerrors.UserError(err), "error upgrading charts")
+			return envinfo.Env.Name, fmt.Errorf("error upgrading charts: %w", nitroerrors.UserError(err))
```
bkeroackdsc, line 803: You can't tell whether this is a User Error or not.

My approach would be to do this annotation in `pkg/nitro/metahelm`. If a FailedEnvironment report is returned, then it is a User Error (probably misconfigured chart), otherwise it's a system error.

pkg/nitro/env/env.go
```diff
@@ -793,7 +813,8 @@ func (m *Manager) update(ctx context.Context, rd *models.RepoRevisionData) (envn
 	chartSpan, ctx := tracer.StartSpanFromContext(ctx, "build_and_install_charts")
 	if err = m.CI.BuildAndInstallCharts(ctx, &metahelm.EnvInfo{Env: ne.env, RC: ne.rc}, mcloc); err != nil {
 		chartSpan.Finish(tracer.WithError(err))
-		return "", errors.Wrap(nitroerrors.UserError(err), "error installing charts")
+		return "", fmt.Errorf("error installing charts: %w", nitroerrors.UserError(err))
```
bkeroackdsc, line 816: Same.

pkg/nitro/errors/errors.go
```diff
@@ -2,8 +2,42 @@ package errors
...
+	meta.ErrGetFileContents,
+	meta.ErrGetDirectoryContents,
+	meta.ErrGetRepoBranches,
+	meta.ErrGetRefForRepo,
+	meta.ErrCreatingDirectory,
+	meta.ErrCreatingSymlink,
+	meta.ErrCreatingFile,
+	meta.ErrWritingFile,
+	meta.ErrShortWrite,
+}
```
bkeroackdsc, line 19: There's no need to enumerate system errors.

pkg/nitro/errors/errors.go
```diff
@@ -2,8 +2,42 @@ package errors
...
+	meta.ErrMissingPath,
+	meta.ErrMultipleDependency,
+	meta.ErrDuplicateDependency,
+	meta.ErrBranchAndRepo,
+	meta.ErrValidateDependencyNames,
+	meta.ErrMalformedValueOverride,
+	meta.ErrMalformedRepoPath,
+	meta.ErrNilLocation,
+	meta.ErrContextCancelled,
+}
```
bkeroackdsc, line 37: There's no need for this enumeration because any named error constant is by definition a User Error.

pkg/nitro/errors/errors.go
```diff
@@ -44,6 +78,11 @@ func CancelledError(err error) error {
 
 // IsUserError finds the first nitro error in the chain and returns true if it is a user error.
 func IsUserError(err error) bool {
+	for _, uErr := range userErrors {
+		if errors.Is(err, uErr) {
+			return true
+		}
+	}
```
bkeroackdsc, line 85: It would be more efficient to have a named error type and then check for type equality.
```go
uerr, ok := err.(*UserError)
return ok
```

pkg/nitro/meta/meta.go
```diff
@@ -44,6 +43,39 @@ func log(ctx context.Context, msg string, args ...interface{}) {
...
+	ErrMissingPath             = errors.New("repo dependency lacks one of ChartPath or ChartRepoPath")
+	ErrMultipleDependency      = errors.New("only one of Repo, ChartPath, or ChartRepoPath may be used")
+	ErrDuplicateDependency     = errors.New("duplicate repository dependency (check for circular dependency declarations)")
+	ErrBranchAndRepo           = errors.New("branch matching and default branch not available if ChartPath or ChartRepoPath is used")
+	ErrValidateDependencyNames = errors.New("error validating dependency names")
+	ErrMalformedValueOverride  = errors.New("malformed value override")
+	ErrMalformedRepoPath       = errors.New("malformed repo path")
+	ErrNilLocation             = errors.New("location is nil")
+	ErrContextCancelled        = errors.New("context was cancelled")
+)
```
bkeroackdsc, line 63: Nice 👍

pkg/nitro/meta/meta.go
```diff
@@ -44,6 +43,39 @@ func log(ctx context.Context, msg string, args ...interface{}) {
...
+
+// System errors
+var (
```
bkeroackdsc, line 66: Unnecessary.

pkg/nitro/meta/meta.go
```diff
@@ -54,13 +86,13 @@ func (g DataGetter) GetAcylYAML(ctx context.Context, rc *models.RepoConfig, repo
 	log(ctx, "fetching acyl.yml for %v@%v", repo, ref)
 	b, err := g.RC.GetFileContents(ctx, repo, "acyl.yml", ref)
 	if err != nil {
-		return errors.Wrap(err, "error getting acyl.yml")
+		return fmt.Errorf("error getting %v@%v: %v: %w", repo, ref, "acyl.yml", ErrGetFileContents)
 	}
 	if err := yaml.Unmarshal(b, &rc); err != nil {
-		return errors.Wrap(err, "error unmarshaling acyl.yml")
+		return ErrYAMLSyntax
```
bkeroackdsc, line 92: You need to wrap the original error so that information about the syntax error isn't lost.

bkeroackdsc (changes requested):
My main issue is that you don't need to explicitly annotate errors as System Errors. Any error is implicitly a system error if it is not annotated explicitly as a User Error.

Also, I think we should take this opportunity to update to Go 1.15 and make sure all our images are using consistent versions (Dockerfile, CircleCI).

https://github.com/dollarshaveclub/acyl/blob/71d3feffc75c78738eedd3b3bd9f8d576c10904f/.circleci/config.yml#L6
https://github.com/dollarshaveclub/acyl/blob/71d3feffc75c78738eedd3b3bd9f8d576c10904f/.circleci/config.yml#L16
https://github.com/dollarshaveclub/acyl/blob/71d3feffc75c78738eedd3b3bd9f8d576c10904f/Dockerfile#L1

pkg/nitro/env/env.go
wwalexander: I think that we might not be able to reliably differentiate those types of errors here, as m.CI is an interface, so we'd have to add requirements to the metahelm.Installer interface to return specific error types. Should we just leave this as a system error?

PR #138, opened 2020-10-29
Add and set environment cancellation status

wwalexander:
This PR adds a Cancelled environment status. This status is set when an operation on the environment (Create, Update, or Delete) is cancelled via a context.


pkg/nitro/env/env.go
```diff
@@ -274,7 +274,16 @@ func (m *Manager) Create(ctx context.Context, rd models.RepoRevisionData) (strin
 		name, err = m.create(ctx, &rd)
 		return err
 	})
-	return name, err
+	cancelled := nitroerrors.IsCancelledError(err)
+	if cancelled {
```
bkeroackdsc, line 278: `if nitroerrors.IsCancelledError(err) {`

pkg/nitro/env/env.go
```diff
@@ -274,7 +274,16 @@ func (m *Manager) Create(ctx context.Context, rd models.RepoRevisionData) (strin
...
+		if statusErr := m.setStatus(ctx, rd, models.Cancelled); statusErr != nil {
+			err = fmt.Errorf("error setting environment status after cancellation: %w", err)
+		}
+	}
+	if err != nil {
+		return "", err
+	}
+	return name, nil
```
bkeroackdsc, line 286: `return name, err`

pkg/nitro/env/env.go
```diff
@@ -553,6 +562,12 @@ func (m *Manager) Delete(ctx context.Context, rd *models.RepoRevisionData, reaso
 	err = m.lockingOperation(ctx, rd.Repo, rd.PullRequest, func(ctx context.Context) error {
 		return m.delete(ctx, rd, reason)
 	})
+	cancelled := nitroerrors.IsCancelledError(err)
+	if cancelled {
```
bkeroackdsc, line 566: `if nitroerrors.IsCancelledError(err) {`

pkg/nitro/env/env.go
```diff
@@ -571,6 +586,17 @@ func (m *Manager) getenv(ctx context.Context, rd *models.RepoRevisionData) (*mod
 	return &envs[0], nil
 }
 
+func (m *Manager) setStatus(ctx context.Context, rd models.RepoRevisionData, status models.EnvironmentStatus) error {
```
bkeroackdsc, line 589: This method doesn't seem necessary. I think you can just use `m.DL.SetQAEnvironmentStatus()` in the places where you use this method.

pkg/nitro/env/env.go
```diff
@@ -683,7 +709,16 @@ func (m *Manager) Update(ctx context.Context, rd models.RepoRevisionData) (strin
 		name, err = m.update(ctx, &rd)
 		return err
 	})
-	return name, err
+	cancelled := nitroerrors.IsCancelledError(err)
+	if cancelled {
```
bkeroackdsc, line 713: See previous comment.

pkg/nitro/env/env.go
```diff
@@ -683,7 +709,16 @@ func (m *Manager) Update(ctx context.Context, rd models.RepoRevisionData) (strin
...
+		if statusErr := m.setStatus(ctx, rd, models.Cancelled); statusErr != nil {
+			err = fmt.Errorf("error setting environment status after cancellation: %w", err)
+		}
+	}
+	if err != nil {
+		return "", err
+	}
+	return name, nil
```
bkeroackdsc, line 721: See previous comment.

bkeroackdsc (changes requested):

pkg/nitro/env/env.go
wwalexander: Whoops, had this in from an earlier version where I used the value again and didn't want to unwrap the error stack twice. Fixed by b52daf6.

pkg/nitro/env/env.go
wwalexander: Fixed by b52daf6.

pkg/nitro/env/env.go
wwalexander: Fixed by b52daf6.

pkg/nitro/env/env.go
wwalexander: Fixed by b52daf6.

pkg/nitro/env/env.go
wwalexander: Fixed by b52daf6.

pkg/nitro/env/env.go
```diff
@@ -274,6 +274,15 @@ func (m *Manager) Create(ctx context.Context, rd models.RepoRevisionData) (strin
 		name, err = m.create(ctx, &rd)
 		return err
 	})
+	if nitroerrors.IsCancelledError(err) {
+		env, err := m.getenv(ctx, &rd)
+		if err != nil {
+			err = fmt.Errorf("error getting environment from RepoRevisionData after cancellation: %w", err)
+		}
```
bkeroackdsc, line 281: I think we can safely disregard any error from `m.getenv()` in this code, because there may be instances where an event gets cancelled but there is no extant environment yet (maybe this was the first create event, and it was cancelled before an environment was ever persisted in the database).

Keep in mind also that `ctx` here is guaranteed to be cancelled, so for `getenv()` you'll need to use a new context otherwise the datalayer methods will return errors.

Change to something like:
```go
env, err2 := m.getenv(ctx, &rd)
if err2 == nil {
   // set cancelled status
}
return name, err
```

pkg/nitro/env/env.go
```diff
@@ -274,6 +274,15 @@ func (m *Manager) Create(ctx context.Context, rd models.RepoRevisionData) (strin
...
+		if err := m.DL.SetQAEnvironmentStatus(ctx, env.Name, models.Cancelled); err != nil {
+			err = fmt.Errorf("error persisting cancelled status after cancellation: %w", err)
```
bkeroackdsc, line 283: You should log this (using the event logger) but don't clobber the original error.

pkg/nitro/env/env.go
```diff
@@ -683,6 +701,15 @@ func (m *Manager) Update(ctx context.Context, rd models.RepoRevisionData) (strin
 		name, err = m.update(ctx, &rd)
 		return err
 	})
+	if nitroerrors.IsCancelledError(err) {
+		env, err := m.getenv(ctx, &rd)
```
bkeroackdsc, line 705: See comments above. The same applies here.

bkeroackdsc (changes requested):

pkg/nitro/env/env.go
wwalexander: Should be fixed by c2746e2.

pkg/nitro/env/env.go
wwalexander: Should be fixed by c2746e2.

pkg/nitro/env/env.go
wwalexander: Should be fixed by c2746e2.

pkg/nitro/env/env.go
```diff
@@ -274,6 +274,16 @@ func (m *Manager) Create(ctx context.Context, rd models.RepoRevisionData) (strin
 		name, err = m.create(ctx, &rd)
 		return err
 	})
+	if nitroerrors.IsCancelledError(err) {
+		ctx := context.Background()
```
bkeroackdsc, line 278: Don't shadow/overwrite `ctx` because the existing context contains the eventlogger that you need below. Just do `env, envErr := m.getenv(context.Background(), &rd)`

pkg/nitro/env/env.go
```diff
@@ -274,6 +274,16 @@ func (m *Manager) Create(ctx context.Context, rd models.RepoRevisionData) (strin
...
+		env, envErr := m.getenv(ctx, &rd)
+		if envErr != nil {
+			return name, err
+		}
+		if err := m.DL.SetQAEnvironmentStatus(ctx, env.Name, models.Cancelled); err != nil {
+			m.log(ctx, "error persisting cancelled status after cancellation: %v", err)
```
bkeroackdsc, line 284: This wouldn't work (would get logged to a null eventlogger) because you've shadowed `ctx` above.

bkeroackdsc (changes requested):

pkg/nitro/env/env.go
wwalexander: Fixed by 4a69411.

pkg/nitro/env/env.go
wwalexander: Fixed by 4a69411.

pkg/nitro/env/env.go
```diff
@@ -274,6 +274,15 @@ func (m *Manager) Create(ctx context.Context, rd models.RepoRevisionData) (strin
 		name, err = m.create(ctx, &rd)
 		return err
 	})
+	if nitroerrors.IsCancelledError(err) {
+		env, envErr := m.getenv(context.Background(), &rd)
+		if envErr != nil {
+			return name, err
+		}
+		if err := m.DL.SetQAEnvironmentStatus(ctx, env.Name, models.Cancelled); err != nil {
```
bkeroackdsc, line 282: You'll need to use `context.Background()` here as well (because ctx is cancelled).

bkeroackdsc (changes requested):

pkg/nitro/env/env.go
wwalexander: Whoops, sorry for missing that! Fixed by 5b5acc3.

bkeroackdsc (approved):

PR #139, opened 2020-11-11
Update default tiller image and pull from Docker Hub

wwalexander:
I grepped the codebase for gcr.io/ and the only instance was DefaultTillerImage, which is now set to `helmpack/tiller:v2.14.3`.


jmedberydsc (approved):
Looks good

PR #141, opened 2020-11-13
Use private ECR image for tiller

wwalexander:
The Tiller image name for Nitro is now configurable via the CLI with the `--tiller-image` flag and via the `Values.app.tiller_image` field via Helm. This flag defaults to `helmpack/tiller:v2.17.0` from Docker Hub, and we will set the image to the private ECR copy in helm-charts to avoid rate-limiting. The integration test uses Docker Hub as we'd have to be running it over 80 times per hour unauthenticated to hit the limit.


jmedberydsc (approved):
looks good.

bkeroackdsc (approved):

PR #143, opened 2020-11-18
Don't report CancelledErrors as errors to Datadog

wwalexander:
This PR avoids reporting CancelledErrors as errors via span.Finish to the webhook_processor span. I think this could also be implemented as adding a tag like `is_cancelled` and then filtering the metric based on the value of that tag, if we want to preserve info on when cancellations have occurred.


pkg/api/api_v0.go
```diff
@@ -213,15 +214,23 @@ func (api *v0api) processWebhook(ctx context.Context, action string, rrd models.
 	setTagsForGithubWebhookHandler(span, rrd)
 	ctx = tracer.ContextWithSpan(ctx, span)
 
+	var err error
+	finishWithError := func() {
+		if nitroerrors.IsCancelledError(err) {
+			span.Finish()
+			return
+		}
+		span.Finish(tracer.WithError(err))
```
bkeroackdsc, line 223: Offhand I think `tracer.WithError(nil)` is the same as `span.Finish()` so this can be simplified.

bkeroackdsc (changes requested):

bkeroackdsc (approved):

esterlinej (commented):
lgtm

PR #149, opened 2020-12-08
Use models.EventLog....WithStatus methods

wwalexander:
This should fix an issue where the EventLog's Status field wasn't getting set correctly by the Postgres datalayer methods. Methods had been added to the models.EventLog that add the status column, but they weren't being used by the datalayer and thus the Status field stayed at its zero value.


esterlinej (approved):
Also ran functional tests with this, build from DQA, the eventlog is providing the updated status with config

esterlinej (approved):
LGTM, again functional test performed on DQA, looks good

bkeroackdsc (approved):

PR #151, opened 2020-12-15
Add zero value default for K8sPod.Restarts

esterlinej:
## Issue
Occured while testing on local build in minikube:

`2020/12/15 09:27:33 http: panic serving [::1]:64366: runtime error: index out of range [0] with length 0
goroutine 119 [running]:
net/http.(*conn).serve.func1(0xc001955ea0)
	/usr/local/go/src/net/http/server.go:1801 +0x147
panic(0x2df2940, 0xc0006c5080)
	/usr/local/go/src/runtime/panic.go:975 +0x3e9
github.com/dollarshaveclub/acyl/pkg/nitro/metahelm.ChartInstaller.GetPodList(0x328d140, 0xc0014bc740, 0x330e800, 0xc00023eb40, 0xc000ddce00, 0x0, 0x30b1888, 0xa, 0x77359400, 0x2f895c0, ...)
	/Users/jesterline/go/src/github.com/dollarshaveclub/acyl/pkg/nitro/metahelm/metahelm.go:1132 +0x679
github.com/dollarshaveclub/acyl/pkg/api.(*v2api).userEnvNamePodsHandler(0xc000de5c00, 0x32d6440, 0xc00141f900, 0xc0025ab900)
	/Users/jesterline/go/src/github.com/dollarshaveclub/acyl/pkg/api/api_v2.go:900 +0x607
github.com/dollarshaveclub/acyl/pkg/api.(*sessionAuthenticator).sessionAuth.func2(0x32d6440, 0xc00141f900, 0xc0025ab800)
	/Users/jesterline/go/src/github.com/dollarshaveclub/acyl/pkg/api/middleware.go:212 +0x2fb
net/http.HandlerFunc.ServeHTTP(0xc00194b5e0, 0x32d6440, 0xc00141f900, 0xc0025ab800)
	/usr/local/go/src/net/http/server.go:2042 +0x44`

## Fix
Default zero value for K8sPod.Restarts, overridden with the k8s `ContainerStatuses[0].RestartCount` if returned.

### Files Updated
- metahelm/metahelm
- metahelm/fake
- metahelm/metahelm_test


wwalexander:
LGTM, but it looks like tests are failing - do we know why? https://app.circleci.com/pipelines/github/dollarshaveclub/acyl/880/workflows/9bc9aff1-5121-4e61-9f11-395dfa616213/jobs/2699

wwalexander (approved):

esterlinej:
The only way I found to properly nil check these k8s resources is to call `.Size()`, which returns 0 if nil

esterlinej:
@wwalexander uncertain, though not related, will dig into and append to this PR though if repeatable

esterlinej:
Fixed the checks on the failing test for APIv2, missed updating this after expanding the dummy pod data

pkg/nitro/metahelm/metahelm.go
```diff
@@ -1125,11 +1125,15 @@ func (ci ChartInstaller) GetPodList(ctx context.Context, ns string) (out []K8sPo
 				nReady += 1
 			}
 		}
+		rc := int32(0)
```
wwalexander, line 1128: Do we need to add the same checks here as fake.go?

esterlinej: Change your view to show all commits, that's probably why

wwalexander: Wow, what a relief! Thanks for the tip.

pkg/nitro/metahelm/metahelm.go
```diff
@@ -1125,11 +1125,19 @@ func (ci ChartInstaller) GetPodList(ctx context.Context, ns string) (out []K8sPo
...
+		if p.Status.Size() != 0 {
```
wwalexander, line 1129: I just figured out that this can be done without a bunch of nested conditionals using switch statements - [Playground](https://play.golang.org/p/9OVZwcJk08F)

In this case, you could do something like:

```go
var rc int32
switch {
case p.Status.Size() == 0:
    fallthrough
case len(p.Status.ContainerStatuses) == 0:
    fallthrough
case p.Status.ContainerStatuses[0].Size() == 0:
    rc = 0
default:
    rc = p.Status.ContainerStatues[0].RestartCount
}
// Use rc
```

If any of the preliminary checks fail, they fallthrough to the last case, which sets the default, and don't check any of the other conditions, avoiding nil pointer dereferences/invalid indexes. If none of the checks fail, the switch statement calls the default expression. Not sure if this is more idiomatic but it seems better to me.

esterlinej: I'm going to update the nested with a single check using `&&`

bkeroackdsc (approved):

wwalexander:
LGTM

PR #153, opened 2021-01-13
Implement User API Keys

esterlinej:
## Requirements
- Implement Middleware Token Authorization per Handler
- Remove deprecated v0 /success, /failure, /event endpoints and handlers
#### Moved to UI API Keys
- ~In the user settings view, implement the ability to create and delete API tokens~
- ~Users should be able to name them with an optional description~
- ~Users should see a list of their currently-active API tokens (names/descriptions) with the ability to delete them as needed.~

---
#### Middleware
- Add context keys:
-- `apiKeyCtxKey` (Authenticated)
-- `qaEnvCtxKey` (QAEnvironment)
-- `eventLogCtxKey` (EventLog)
- Add `tokenAuth()` writes `Authenticated` struct with `IsAdmin` (bool) or `User` (string) to context for use with authorizors
- Add authorizors (`authorize()`, `authorizeEnv()`, `authorizeEventLog()`) in middleware chain, called after and requiring tokenAuth
#### API
- Remove deprecated `v0/envs/{name}/success`, `/failure`, `/event`
- Update API non-versioned/v0/v1/v2 paths & handlers that use `tokenAuth` to support changes, and updated or new tests
#### Tokens
- Removed, coming with UI Implementation
- ~Token handlers verify Session before proceeding, confirming user creating token~
- ~Token path added: `GET /v2/user/tokens` returns list of tokens~
- ~Token path added: `POST /v2/user/token` creates user token~ 
- ~Token path added: `DELETE /v2/user/token/{id}` deletes user token~
#### Persistence
- Added `GetMostRecentForUser()`, `SearchEnvsForUser()` to the datalayer, fakes
- Added db tests for new functions


pkg/api/api_v0.go
```diff
@@ -415,7 +415,18 @@ func (api *v0api) envDetailHandler(w http.ResponseWriter, r *http.Request) {
 		api.notfoundError(w)
 		return
 	}
-
+	if !r.Context().Value(acylAPIKey).(bool) {
```
wwalexander, line 418: I think this could potentially be simplified to:

```
key, ok := r.Context().Value(userAPIKey).(models.APIKey)
if !ok || key.GitHubUser != qa.User {
    api.forbiddenError(w, "request not authorized")
    return
}
```

pkg/api/api_v2.go
```diff
@@ -1085,3 +1122,208 @@ func (api *v2api) userEnvPodLogsHandler(w http.ResponseWriter, r *http.Request)
...
+		return
+	}
+	v2uakc := V2UserAPIKeyCreate{
+		User:  user,
+		Token: keyID,
+	}
+	w.Header().Set("Content-Type", "application/json")
+	if err := json.NewEncoder(w).Encode(&v2uakc); err != nil {
+		api.rlogger(r).Logf("error marshaling new user token: %v", err)
+		w.WriteHeader(http.StatusInternalServerError)
```
wwalexander, line 1185: Writing a header after starting the body isn't supported by HTTP AFAIK; you could buffer the JSON if you need to be able to detect encoding failures or use [trailers](https://golang.org/pkg/net/http/#example_ResponseWriter_trailers)

pkg/api/api_v0.go
esterlinej: I will look into this, the issue is we need to let Admin (acyl api key) go through if true, otherwise we check for user api key.

pkg/api/api_v2.go
esterlinej: So review all other handlers, this is done the exact same way.

wwalexander (approved):
LGTM besides for the couple of random nitpicks I had.

pkg/api/api_v0.go
wwalexander: Ah yeah, I left that comment in a misleading spot, I was thinking it would go inside the check for the acylAPIKey, e.g.

```
if !r.context().Value(acylAPIKey).(bool) {
    key, ok := r.Context().Value(userAPIKey).(models.APIKey)
    if !ok || key.GitHubUser != qa.User {
        api.forbiddenError(w, "request not authorized")
        return
    }
}
```

esterlinej: So given that we need to first check for acyl api key, and continue if found, we'd need to wrap it to account for both options, cause if false and no user key exists or doesnt match, it needs to error for this as well. Using your suggestion, I have rewritten as such:
```
if !r.Context().Value(acylAPIKey).(bool) {
	var authorized bool
	key, ok := r.Context().Value(userAPIKey).(models.APIKey)
	if ok && key.GitHubUser == qa.User {
		authorized = true
	}
	if !authorized {
		api.forbiddenError(w, "request not authorized")
		return
	}
}
```

wwalexander:
```
var authorized bool
key, ok := r.Context().Value(userAPIKey).(models.APIKey)
if ok && key.GitHubUser == qa.User {
    authorized = true
}
if !authorized {
    ...
}
```

is equivalent to

```
key, ok := r.Context().Value(userAPIKey).(models.APIKey)
authorized = ok && key.GitHubUser == qa.User
if !authorized {
    ...
}
```

is equivalent to 

```
key, ok := r.Context().Value(userAPIKey).(models.APIKey)
authorized = ok && key.GitHubUser == qa.User
if !(ok && key.GitHubUser == qa.User) {
    ...
}
```

is equivalent to 

```
key, ok := r.Context().Value(userAPIKey).(models.APIKey)
authorized = ok && key.GitHubUser == qa.User
if !ok || key.GitHubUser != qa.User {
    ...
}
```

(by De Morgans' theorem) so these should be the same. But it's a really minor nitpick and maybe it's less readable.

pkg/api/api_v2.go
wwalexander: True, the only side effect (besides that the client will only see a 200 status code) is that the default server seems to log a message about `superfluous response.WriteHeader call` when it happens but it doesn't seem to cause any actual issues.

pkg/api/api_v0.go
esterlinej: Okay, completely reworked. Single authorized apibase func, and more testing to check authorized/unauthorized for user api keys.

esterlinej:
@wwalexander Okay, completely reworked. Single authorized apibase func, and more testing to check authorized/unauthorized for user api keys.

pkg/api/api_v2.go
esterlinej: We'll talk in depth on this, make sure we're on the same page.

pkg/api/api_v0.go
esterlinej: See notes, completely reworked.

pkg/api/api_v2.go
bkeroackdsc: Good catch @wwalexander. Other handlers making the same error is not an excuse to continue doing it.

pkg/api/api.go
```diff
@@ -49,6 +51,20 @@ func (api *apiBase) forbiddenError(w http.ResponseWriter, msg string) {
...
 
+// authorized first checks for acyl then user api key
+// if acyl api key is:
+// - true: returns validated bypassing the user check
+// - false: checks for and validates user on user api key to confirm authorization
+func (api *apiBase) authorized(ctx context.Context, user string) bool {
+	if auth, ok := ctx.Value(acylAPIKey).(bool); ok && auth {
+		return true
+	}
+	if key, ok := ctx.Value(userAPIKey).(*models.APIKey); ok && key.GitHubUser == user {
```
bkeroackdsc, line 62: Don't type assert twice. Get the value once from the context, then perform your logic.

pkg/api/middleware.go
```diff
@@ -30,6 +30,8 @@ func middlewareChain(f http.HandlerFunc, m ...middleware) http.HandlerFunc {
 
 const (
 	apiKeyHeader = "API-Key"
+	acylAPIKey   = "api_key"
+	userAPIKey   = "user_api_key"
```
bkeroackdsc, line 34: These names should refer to context keys. The word "acyl" is redundant also. For example, `apiKeyCtxKey`.

I don't see why we need separate "user api" key and "api key" context keys.

pkg/api/api_v2.go
```diff
@@ -1085,3 +1106,208 @@ func (api *v2api) userEnvPodLogsHandler(w http.ResponseWriter, r *http.Request)
 		w.WriteHeader(http.StatusInternalServerError)
 	}
 }
+
+type V2UserAPIKeyCreate struct {
+	User  string    `json:"user"`
+	Token uuid.UUID `json:"token"`
+}
```
bkeroackdsc, line 1113: Don't let the user specify the UUID. We should generate a random one internally and then return it.

pkg/api/api_v2.go
```diff
@@ -1085,3 +1106,208 @@ func (api *v2api) userEnvPodLogsHandler(w http.ResponseWriter, r *http.Request)
...
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+	switch perm {
+	case "1":
+		key.PermissionLevel = models.ReadOnlyPermission
+	case "2":
+		key.PermissionLevel = models.WritePermission
+	case "3":
+		key.PermissionLevel = models.AdminPermission
```
bkeroackdsc, line 1146: Why are we using string integers query parameters? This would be better modeled with a field in the JSON request payload.

pkg/api/api_v2.go
```diff
@@ -1085,3 +1106,208 @@ func (api *v2api) userEnvPodLogsHandler(w http.ResponseWriter, r *http.Request)
...
+	}
+	key.Name = r.URL.Query().Get("name")
```
bkeroackdsc, line 1148: Why a query param? You have a JSON request payload.

pkg/api/middleware.go
```diff
@@ -73,7 +75,7 @@ func (ra reqAuthorizor) tokenAuth(f http.HandlerFunc, minPermission models.Permi
 						w.WriteHeader(http.StatusInternalServerError)
 						return
 					}
-					f(w, r)
+					f(w, r.Clone(context.WithValue(r.Context(), userAPIKey, apikey)))
```
bkeroackdsc, line 78: I don't think you need to have different context keys for "user" keys vs admin/root keys.

pkg/api/api_v2.go
```diff
@@ -1085,3 +1106,208 @@ func (api *v2api) userEnvPodLogsHandler(w http.ResponseWriter, r *http.Request)
...
+		api.rlogger(r).Logf("action not authorized by user: %v", uis.GitHubUser)
+		w.WriteHeader(http.StatusUnauthorized)
+		return
+	}
+	go func() {
+		err := api.dl.DeleteAPIKey(context.Background(), keyID)
+		if err != nil {
+			api.logger.Printf("error deleting user token: %v: %v", id, err)
+		}
+	}()
```
bkeroackdsc, line 1311: This doesn't need to be asynchronous.

pkg/api/api_v2.go
```diff
@@ -1085,3 +1106,208 @@ func (api *v2api) userEnvPodLogsHandler(w http.ResponseWriter, r *http.Request)
...
+		w.WriteHeader(http.StatusInternalServerError)
+	}
+}
+
+func (api *v2api) apiKeyDestroyHandler(w http.ResponseWriter, r *http.Request) {
+	uis, err := getSessionFromContext(r.Context())
+	if err != nil {
+		api.rlogger(r).Logf("session missing from context")
+		w.WriteHeader(http.StatusBadRequest)
+		return
```
bkeroackdsc, line 1281: Why are you getting the UI session? This is unneeded.

bkeroackdsc (changes requested):

pkg/api/api_v2.go
esterlinej: Will and I dug into this, it stems from an old issue with 1.11, and he confirmed it's not an issue in playground.

pkg/api/middleware.go
esterlinej: We need a way to check the root vs user key.

esterlinej: Didn't realize I could simply piggyback off the same key at the time.

pkg/api/api_v2.go
esterlinej: They don't specify the key, this is the return object only

pkg/api/api_v2.go
esterlinej: This was to confirm the user in session making the request to create/get/destroy the key

pkg/api/middleware.go
bkeroackdsc: One context key should always map to one data structure in the context. You can then use fields within that data structure in your logic.

pkg/api/api_v2.go
bkeroackdsc: Use different struct types for the request and response schemas.

esterlinej: I see, the naming is confusing, sorry, will update this.

pkg/api/middleware.go
```diff
@@ -82,6 +87,147 @@ func (ra reqAuthorizor) tokenAuth(f http.HandlerFunc, minPermission models.Permi
...
+	return func(w http.ResponseWriter, r *http.Request) {
+		log := func(msg string, args ...interface{}) {
+			log.Printf("authorizedEnv: "+msg, args...)
+		}
+		ak := r.Context().Value(apiKeyCtxKey)
+		if auth, ok := ak.(bool); ok && auth {
+			f(w, r)
+			return
+		}
+		user := r.URL.Query().Get("user")
```
esterlinej, line 177: The `env/_search/` endpoints will reject forbidden 403 if a user api key is used without searching with their own name, I think we should handle this gracefully rather. Thoughts?

pkg/api/api_v0.go
```diff
@@ -140,9 +140,9 @@ func (api *v0api) register(r *muxtrace.Router) error {
...
 	r.HandleFunc("/v0/webhook", middlewareChain(api.legacyGithubWebhookHandler, waitMiddleware.waitOnRequest)).Methods("POST")
-	r.HandleFunc("/v0/envs/_search", middlewareChain(authMiddleware.tokenAuth(api.envSearchHandler, models.ReadOnlyPermission))).Methods("GET")
-	r.HandleFunc("/v0/envs/{name}", middlewareChain(authMiddleware.tokenAuth(api.envDetailHandler, models.ReadOnlyPermission))).Methods("GET")
-	r.HandleFunc("/v0/envs/{name}", middlewareChain(authMiddleware.tokenAuth(api.envDestroyHandler, models.AdminPermission), waitMiddleware.waitOnRequest)).Methods("DELETE")
+	r.HandleFunc("/v0/envs/_search", middlewareChain(authMiddleware.tokenAuth(authMiddleware.authorizedSearch(api.envSearchHandler), models.ReadOnlyPermission))).Methods("GET")
+	r.HandleFunc("/v0/envs/{name}", middlewareChain(authMiddleware.tokenAuth(authMiddleware.authorizedEnv(api.envDetailHandler), models.ReadOnlyPermission))).Methods("GET")
+	r.HandleFunc("/v0/envs/{name}", middlewareChain(authMiddleware.tokenAuth(authMiddleware.authorizedEnv(api.envDestroyHandler), models.AdminPermission), waitMiddleware.waitOnRequest)).Methods("DELETE")
 
 	// DEPRECATED: no longer supported, please use /v0/envs/_search
```
esterlinej, line 147: Deprecated endpoints that go through tokenAuth are not updated with this PR, and provide end-arounds to access the data. I recommend we remove these deprecated endpoints rather than provide support.

pkg/api/api_v2.go
```diff
@@ -420,21 +416,7 @@ func (api *v2api) healthCheck(w http.ResponseWriter, r *http.Request) {
...
-	el, err := api.dl.GetEventLogByID(id)
-	if err != nil {
-		api.internalError(w, errors.Wrap(err, "error fetching event logs"))
-		return
-	}
-	if el == nil {
-		api.notfoundError(w)
-		return
-	}
+	el := r.Context().Value(dataCtx).(*models.EventLog)
```
bkeroackdsc, line 419: Value not pointer. Use a different context key.

pkg/api/api_v2.go
```diff
@@ -1085,3 +1067,206 @@ func (api *v2api) userEnvPodLogsHandler(w http.ResponseWriter, r *http.Request)
 		w.WriteHeader(http.StatusInternalServerError)
 	}
 }
+
+type v2UserAPIKeyCreate struct {
```
bkeroackdsc, line 1071: Name these types to indicate if they are request or response structs.

pkg/api/api_v2.go
```diff
@@ -1085,3 +1067,206 @@ func (api *v2api) userEnvPodLogsHandler(w http.ResponseWriter, r *http.Request)
...
+	if err != nil {
+		api.rlogger(r).Logf("session missing from context")
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+	key := v2UserAPIKeyCreate{}
+	err = json.NewDecoder(r.Body).Decode(&key)
+	if err != nil {
+		api.rlogger(r).Logf("error unmarshaling body: %v", err)
+		w.WriteHeader(http.StatusInternalServerError)
```
bkeroackdsc, line 1094: This should be 403 Bad Request

pkg/api/api_v2.go
```diff
@@ -1085,3 +1067,206 @@ func (api *v2api) userEnvPodLogsHandler(w http.ResponseWriter, r *http.Request)
 		w.WriteHeader(http.StatusInternalServerError)
 	}
 }
+
+type v2UserAPIKeyCreate struct {
+	Permission  int    `json:"permission"`
+	Name        string `json:"name"`
+	Description string `json:"description"`
+	User        string `json:"user"`
```
bkeroackdsc, line 1075: Don't allow the client to specify user. User should be inferred by the credentials used to authorize the API call.

pkg/api/api_v2.go
```diff
@@ -1085,3 +1067,206 @@ func (api *v2api) userEnvPodLogsHandler(w http.ResponseWriter, r *http.Request)
...
+		api.rlogger(r).Logf("action not authorized by user: %v", uis.GitHubUser)
+		w.WriteHeader(http.StatusUnauthorized)
+		return
+	}
+	if key.User == "" {
+		api.rlogger(r).Logf("error token user required")
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+	if key.Permission < models.MinPermissionLevel || key.Permission > models.MaxPermissionLevel {
```
bkeroackdsc, line 1107: You may want to explicitly check for `key.Permission < 0` (since the field is a signed integer).

pkg/api/api_v2.go
```diff
@@ -1085,3 +1067,206 @@ func (api *v2api) userEnvPodLogsHandler(w http.ResponseWriter, r *http.Request)
...
+	}
+	w.Header().Set("Content-Type", "application/json")
+	if err := json.NewEncoder(w).Encode(&v2uakc); err != nil {
+		api.rlogger(r).Logf("error marshaling new user token: %v", err)
+		w.WriteHeader(http.StatusInternalServerError)
+	}
+}
+
+type V2UserAPIKey struct {
+	Token       uuid.UUID              `json:"token"`
```
bkeroackdsc, line 1135: Never return the actual token. That should be displayed to the user once on creation and then never again.

pkg/api/api_v2.go
```diff
@@ -1085,3 +1067,206 @@ func (api *v2api) userEnvPodLogsHandler(w http.ResponseWriter, r *http.Request)
...
+	Created     time.Time              `json:"created"`
+	LastUsed    pq.NullTime            `json:"last_used"`
+	Permission  models.PermissionLevel `json:"permission"`
+	Name        string                 `json:"name"`
+	Description string                 `json:"description"`
+	User        string                 `json:"user"`
```
bkeroackdsc, line 1141: The user should be inferred by the credentials used to authorize this API call. We can have a separate API endpoint authorized by root API tokens which allows displaying all API tokens (or by user, etc). I'm not sure if we need that yet however.

pkg/api/api_v2.go
```diff
@@ -1085,3 +1067,206 @@ func (api *v2api) userEnvPodLogsHandler(w http.ResponseWriter, r *http.Request)
...
+	}
+	if len(apikeys) == 0 {
+		w.WriteHeader(http.StatusNotFound)
+		return
+	}
+	if uis.GitHubUser != apikeys[0].GitHubUser {
+		api.rlogger(r).Logf("action not authorized by user: %v", uis.GitHubUser)
+		w.WriteHeader(http.StatusUnauthorized)
+		return
+	}
```
bkeroackdsc, line 1171: Always perform security/authorization checks as early as possible in the handler. This can be put before you query for keys by user.

pkg/api/api_v2.go
```diff
@@ -1085,3 +1067,206 @@ func (api *v2api) userEnvPodLogsHandler(w http.ResponseWriter, r *http.Request)
...
+	v2uaks := []V2UserAPIKey{}
```
bkeroackdsc, line 1172: Since you know the length of `apikeys` you can initialize this with a size: `v2uaks := make([]V2UserAPIKey, len(apikeys))`

pkg/api/api_v2.go
```diff
@@ -1085,3 +1067,206 @@ func (api *v2api) userEnvPodLogsHandler(w http.ResponseWriter, r *http.Request)
...
+		})
+	}
+	w.Header().Set("Content-Type", "application/json")
+	if err := json.NewEncoder(w).Encode(&v2uaks); err != nil {
+		api.rlogger(r).Logf("error marshaling new user token: %v", err)
+		w.WriteHeader(http.StatusInternalServerError)
+	}
+}
+
+func (api *v2api) apiKeyByIdHandler(w http.ResponseWriter, r *http.Request) {
```
bkeroackdsc, line 1191: I don't think we need this endpoint.

pkg/api/middleware.go
```diff
@@ -82,6 +87,147 @@ func (ra reqAuthorizor) tokenAuth(f http.HandlerFunc, minPermission models.Permi
...
+		if err != nil {
+			log(errors.Wrap(err, "error fetching event logs").Error())
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+		if el == nil {
+			w.WriteHeader(http.StatusNotFound)
+			return
+		}
+		qa, err := ra.dl.GetQAEnvironmentConsistently(r.Context(), el.EnvName)
```
bkeroackdsc, line 148: There's no difference between `GetQAEnvironmentConsistently()` and `GetQAEnvironment()`. This is tech debt from when DynamoDB was used.

pkg/api/middleware.go
```diff
@@ -82,6 +87,147 @@ func (ra reqAuthorizor) tokenAuth(f http.HandlerFunc, minPermission models.Permi
...
+	if auth, ok := ak.(bool); ok && auth {
+		return true
+	}
+	if key, ok := ak.(*models.APIKey); ok && key.GitHubUser == user {
+		return true
+	}
+	return false
+}
+
+func (ra reqAuthorizor) authorizedEnv(f http.HandlerFunc) http.HandlerFunc {
```
bkeroackdsc, line 101: Document these middleware methods both with what they do, and also the assumptions/expectations they have. For example, they assume that `tokenAuth()` middleware is used prior to them in the chain and will fail if that isn't the case.

pkg/persistence/pg.go
```diff
@@ -501,6 +501,16 @@ func (p *PGLayer) GetMostRecent(ctx context.Context, n uint) ([]QAEnvironment, e
...
 }
 
+// GetMostRecent finds the most recent environments for user from the last n days.
+// Recency is defined by created/updated timestamps.
+// The returned slice is in descending order of recency.
+func (p *PGLayer) GetMostRecentForUser(ctx context.Context, user string, n uint) ([]QAEnvironment, error) {
+	if isCancelled(ctx) {
+		return nil, errors.Wrap(ctx.Err(), "error getting most recent")
+	}
+	return p.collectRows(p.db.QueryContext(ctx, `SELECT `+models.QAEnvironment{}.Columns()+` from qa_environments WHERE username = $1 AND `+fmt.Sprintf(`created >= (current_timestamp - interval '%v days')`, n)+` ORDER BY created DESC;`, user))
```
bkeroackdsc, line 511: You're missing a `LIMIT` clause (using the n method parameter).

bkeroackdsc (changes requested):

pkg/api/middleware.go
bkeroackdsc: I would implement a new datalayer method `SearchEnvsForUser` which adds a mandatory user to the search parameters.

pkg/api/api_v0.go
```diff
@@ -405,17 +405,7 @@ func (api *v0api) envListHandler(w http.ResponseWriter, r *http.Request) {
...
-	if err != nil {
-		api.internalError(w, fmt.Errorf("error getting environment: %v", err))
-		return
-	}
-	if qa == nil {
-		api.notfoundError(w)
-		return
-	}
-
+	qa := r.Context().Value(dataCtx).(*models.QAEnvironment)
```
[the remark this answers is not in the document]
esterlinej: I will try again, but wasn't working until I made this a pointer

pkg/api/api_v2.go
esterlinej: Makes sense, I was thinking about this through more from a user controlled "I need a token to do x with these permissions" but if they don't have those permissions to begin with, then this breaks

pkg/api/api_v2.go
esterlinej: Won't be able to move this up, I removed user from the field and forgot to clean it up out of the handler. The only user we have is the `uis.GitHubUser` prior to fetching the key, then confirming that the user on the key id matches.

pkg/persistence/pg.go
esterlinej: Looking into this, I duplicated the existing method without user, and searched options to confirm this was still the right way to do this, and didn't have `LIMIT` included:
https://stackoverflow.com/questions/5465484/how-to-list-records-with-date-from-the-last-10-days

pkg/api/api_v2.go
esterlinej: I'm now pulling user from the session

pkg/api/api_v0.go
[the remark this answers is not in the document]
esterlinej: I'll add nil checks, cause the bigger issue is we have to then convert this to a copy, then convert it back to a pointer for the handler to use.

esterlinej:
We may want/need to add a delete key by name, in case we need to cleanup keys without ids?

pkg/persistence/pg.go
esterlinej: Resolving per our conversation

pkg/api/api_v0.go
```diff
@@ -365,12 +354,22 @@ func (api *v0api) legacyGithubWebhookHandler(w http.ResponseWriter, r *http.Requ
...
 func (api *v0api) envListHandler(w http.ResponseWriter, r *http.Request) {
-	var fullDetails bool
-	envs, err := api.dl.GetQAEnvironments(r.Context())
+	user, ok := r.Context().Value(userCtxKey).(string)
+	if !ok {
+		api.internalError(w, fmt.Errorf("error authorize context failed"))
+	}
+	envs := []models.QAEnvironment{}
+	var err error
+	if user == acylKey {
```
bkeroackdsc, line 363: Instead of using a string constant as a sentinel to indicate that the API key is an admin key, instead store a struct in the context with a "IsAdmin" boolean field.

pkg/api/api_v0.go
```diff
@@ -405,18 +404,12 @@ func (api *v0api) envListHandler(w http.ResponseWriter, r *http.Request) {
...
-		api.notfoundError(w)
-		return
+	qa, ok := r.Context().Value(qaEnvCtxKey).(models.QAEnvironment)
+	if !ok {
+		api.internalError(w, fmt.Errorf("error authorize context failed"))
 	}
-
-	output := v0QAEnvironmentFromQAEnvironment(qa)
+	qap := &qa
+	output := v0QAEnvironmentFromQAEnvironment(qap)
```
bkeroackdsc, line 412: You don't need the additional named variable `qap`. Do `output := v0QAEnvironmentFromQAEnvironment(&qa)` instead

pkg/api/api_v0.go
```diff
@@ -427,67 +420,25 @@ func (api *v0api) envDetailHandler(w http.ResponseWriter, r *http.Request) {
...
-	if err != nil {
-		api.internalError(w, err)
-		return
-	}
-	if qa == nil {
-		api.notfoundError(w)
-		return
+	qa, ok := r.Context().Value(qaEnvCtxKey).(models.QAEnvironment)
+	if !ok {
+		api.internalError(w, fmt.Errorf("error authorize context failed"))
```
bkeroackdsc, line 425: Use a more informative error message. Something like `fmt.Errorf("unexpected qa env type from context: %T", qa)`

pkg/api/api_v0.go
```diff
@@ -427,67 +420,25 @@ func (api *v0api) envDetailHandler(w http.ResponseWriter, r *http.Request) {
...
-	err = api.dl.AddEvent(r.Context(), name, event.Message)
-	if err != nil {
-		api.internalError(w, err)
-		return
-	}
-	w.WriteHeader(http.StatusNoContent)
-}
-
 func (api *v0api) envSearchHandler(w http.ResponseWriter, r *http.Request) {
+	userSearch, ok := r.Context().Value(searchCtxKey).(bool)
```
bkeroackdsc, line 438: Use the same mechanism here as you use above for determining if this is a user api key or an admin key.

pkg/api/middleware.go
```diff
@@ -29,15 +30,21 @@ func middlewareChain(f http.HandlerFunc, m ...middleware) http.HandlerFunc {
 }
 
 const (
-	apiKeyHeader = "API-Key"
+	apiKeyHeader   = "API-Key"     // string
+	apiKeyCtxKey   = "api_key"     // bool or models.APIKey
```
bkeroackdsc, line 34: A context key must always map to exactly one data type. Don't overload it (bool and models.APIKey).

pkg/api/middleware.go
```diff
@@ -29,15 +30,21 @@ func middlewareChain(f http.HandlerFunc, m ...middleware) http.HandlerFunc {
...
+	acylKey        = "acyl-key"
 )
 
 // authMiddleware checks for correct API key header or aborts with Unauthorized
 var authMiddleware = reqAuthorizor{}
 
 type reqAuthorizor struct {
 	apiKeys []string
-	DL      persistence.APIKeyDataLayer
+	dl      persistence.DataLayer
```
bkeroackdsc, line 47: What's the reasoning behind making this field private?

pkg/api/middleware.go
```diff
@@ -82,6 +89,128 @@ func (ra reqAuthorizor) tokenAuth(f http.HandlerFunc, minPermission models.Permi
...
+	}
+	return false
+}
+
+// authorizedEnv requires tokenAuth to be called first or will fail
+func (ra reqAuthorizor) authorizedEnv(f http.HandlerFunc) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		log := func(msg string, args ...interface{}) {
+			log.Printf("authorizedEnv: "+msg, args...)
+		}
```
bkeroackdsc, line 108: You can put this outside of the closure to avoid allocating a new anonymous function for every handler request.

pkg/api/middleware.go
```diff
@@ -82,6 +89,128 @@ func (ra reqAuthorizor) tokenAuth(f http.HandlerFunc, minPermission models.Permi
...
+			log("request not authorized")
+			w.WriteHeader(http.StatusForbidden)
+			return
+		}
+		f(w, r.Clone(context.WithValue(r.Context(), eventLogCtxKey, *el)))
+	}
+}
+
+// authorizedSearch requires tokenAuth to be called first or will fail
+func (ra reqAuthorizor) authorizedSearch(f http.HandlerFunc) http.HandlerFunc {
```
bkeroackdsc, line 172: It doesn't appear that there's anything specific to search for this middleware, so it should be renamed to something more descriptive.

pkg/api/middleware.go
```diff
@@ -82,6 +89,128 @@ func (ra reqAuthorizor) tokenAuth(f http.HandlerFunc, minPermission models.Permi
...
+			return
+		}
+		log("request not authorized")
+		w.WriteHeader(http.StatusForbidden)
+		return
+	}
+}
+
+// authorizedUser requires tokenAuth to be called first or will fail
+func (ra reqAuthorizor) authorizedUser(f http.HandlerFunc) http.HandlerFunc {
```
bkeroackdsc, line 194: The logic in `authorizedUser()` and `authorizedSearch()` seem to be the same. Combine and rename.

pkg/persistence/pg.go
```diff
@@ -490,6 +490,87 @@ func (p *PGLayer) Search(ctx context.Context, opts models.EnvSearchParameters) (
 	return p.collectRows(p.db.QueryContext(ctx, q, sargs...))
 }
 
+// SearchEnvsForUser finds environments for specified user that satsify the parameters given.
+// Multiple parameters are combined with implicit AND.
+func (p *PGLayer) SearchEnvsForUser(ctx context.Context, opts models.EnvSearchParameters) ([]QAEnvironment, error) {
```
bkeroackdsc, line 495: Don't copy/paste the entire SearchEnvs method logic. Just add the user parameter and then call SearchEnvs.

bkeroackdsc (changes requested):

pkg/api/middleware.go
esterlinej: I had them separate prior and you asked me to combine them. I will revert this change.

pkg/api/middleware.go
esterlinej: My thought behind this is limiting it's scope to the api package.

pkg/api/middleware.go
esterlinej: True, I refactored this, but didn't alter the name, even though its more generic than before.

pkg/api/middleware.go
esterlinej: One returns a string, the other a bool. Let me see what I can do.

pkg/api/api_v0.go
esterlinej: Ha, I did this at first and then thought I was over-engineering it that way.

pkg/api/middleware.go
bkeroackdsc: I would never ask you to overload a context key with multiple types. You must've misunderstood.

pkg/api/api_v0.go
bkeroackdsc: Actually APIKey has a PermissionLevel field you can use to determine if it's admin or not
https://github.com/dollarshaveclub/acyl/blob/3e2477a8df5379239057a29050ffb23f6b476b84/pkg/models/apikeys.go#L16

esterlinej: Sure, though that currently is configurable, and `IsAdmin` to me is about is this a root key vs user key

esterlinej: Added in the `authenticated()` function

pkg/api/api_v2.go
```diff
@@ -420,22 +423,12 @@ func (api *v2api) healthCheck(w http.ResponseWriter, r *http.Request) {
...
-	if el == nil {
-		api.notfoundError(w)
-		return
+	el, ok := r.Context().Value(eventLogCtxKey).(models.EventLog)
+	if !ok {
+		api.internalError(w, fmt.Errorf("error authorize context failed"))
 	}
-	j, err := json.Marshal(v2EventLogFromEventLog(el))
+	elp := &el
+	j, err := json.Marshal(v2EventLogFromEventLog(elp))
```
wwalexander, line 431: Could be shortened to just
`j, err := json.Marshal(v2eventLogFromEventLog(&el))`

pkg/persistence/pg.go
```diff
@@ -501,6 +511,16 @@ func (p *PGLayer) GetMostRecent(ctx context.Context, n uint) ([]QAEnvironment, e
...
 }
 
+// GetMostRecent finds the most recent environments for user from the last n days.
+// Recency is defined by created/updated timestamps.
+// The returned slice is in descending order of recency.
+func (p *PGLayer) GetMostRecentForUser(ctx context.Context, user string, n uint) ([]QAEnvironment, error) {
+	if isCancelled(ctx) {
+		return nil, errors.Wrap(ctx.Err(), "error getting most recent")
+	}
+	return p.collectRows(p.db.QueryContext(ctx, `SELECT `+models.QAEnvironment{}.Columns()+` from qa_environments WHERE username = $1 AND `+fmt.Sprintf(`created >= (current_timestamp - interval '%v days')`, n)+` ORDER BY created DESC;`, user))
```
wwalexander, line 521: While I don't think this particular usage of fmt.Sprintf inside a SQL string could lead to injection, we should probably avoid doing it wherever possible. In this case, you can create an interval from a query parameter like:

    return p.collectRows(p.db.QueryContext(ctx, `SELECT `+models.QAEnvironment{}.Columns()+` from qa_environments WHERE username = $1 AND created >= (current_timestamp - make_interval(days := $2)) ORDER BY created DESC;`, user, n))

wwalexander (approved):
LGTM, besides the couple of minor improvements I suggested.

pkg/api/api_v2.go
esterlinej: Good catch, missed removing this

pkg/persistence/pg.go
esterlinej: Thanks, will change

esterlinej:
The best way I found to do this, have `tokenAuth` write `Authenticated` to the `apiKeyCtxKey` Context, removing the need to the other two, along with the `authenticated()` function, and have each authorizor function get that context to confirm against the Authenticated root key / admin or user.

esterlinej:
Updated PR Description

esterlinej:
Change applied to simplify, but I'm still reviewing to make certain I understand your concerns.

To add, the first `tokenAuth()` simply checks the token is either a valid root or user key, but doesn't otherwise check for authorization, which is where the authorizor functions come into effect. They then ensure the request is authorized before proceeding to the handler.

esterlinej:
Even more simplification, if root admin key is used, it inserts `models.APIKey{PermissionLevel: models.AdminPermission}` into the `apiKeyCtxKey`, otherwise if user it inserts that User API Key

pkg/models/apikeys.go
```diff
@@ -53,3 +53,8 @@ const (
 	WritePermission
 	AdminPermission
 )
+
+const (
+	MinPermissionLevel = 1
+	MaxPermissionLevel = 3
```
bkeroackdsc, line 59: Never rely on the internal literal values of enums since it's an implementation detail of the compiler. Regardless of that, this is unneeded. Use the symbolic constant names instead.

pkg/persistence/pg.go
```diff
@@ -501,6 +511,17 @@ func (p *PGLayer) GetMostRecent(ctx context.Context, n uint) ([]QAEnvironment, e
...
 }
 
+// GetMostRecent finds the most recent environments for user from the last n days.
+// Recency is defined by created/updated timestamps.
+// The returned slice is in descending order of recency.
+func (p *PGLayer) GetMostRecentForUser(ctx context.Context, user string, n uint) ([]QAEnvironment, error) {
+	if isCancelled(ctx) {
+		return nil, errors.Wrap(ctx.Err(), "error getting most recent")
+	}
+	days := fmt.Sprintf(`created >= (current_timestamp - interval '%v days')`, n)
```
bkeroackdsc, line 521: This is a (very remote) SQL injection vulnerability. Never use Sprintf to create SQL queries, always use query placeholders ("$1", "$2"...). For this, use an additional query placeholder in the line below.

bkeroackdsc (changes requested):

pkg/models/apikeys.go
esterlinej: Understood. I forgot to remove this from a prior refactor.

pkg/persistence/pg.go
esterlinej: Unfortunately I was not able to find a better solution since this is wrapped within a string literal, and must be resolved prior. The only real protection is that it's using the `uint`, though I understand the concern.

pkg/api/middleware.go
```diff
@@ -82,6 +86,108 @@ func (ra reqAuthorizor) tokenAuth(f http.HandlerFunc, minPermission models.Permi
...
+			return
+		}
+		if qa == nil {
+			w.WriteHeader(http.StatusNotFound)
+			return
+		}
+		apikey := models.APIKey{}
+		if ak, ok := r.Context().Value(apiKeyCtxKey).(models.APIKey); ok {
+			apikey = ak
+		}
```
bkeroackdsc, line 109: If the APIKey isn't found in the context, abort and return a 403 unauthorized to the client.

```go
ak, ok := r.Context().Value(apiKeyCtxKey).(models.APIKey)
if !ok || ak.User == "" {
   w.WriteHeader(http.StatusForbidden)
   return
}
```

pkg/api/middleware.go
```diff
@@ -82,6 +86,108 @@ func (ra reqAuthorizor) tokenAuth(f http.HandlerFunc, minPermission models.Permi
...
```
bkeroackdsc, line 157: Same here. If APIKey isn't found in the context, return a 403 to the client.

pkg/api/middleware.go
```diff
@@ -82,6 +86,108 @@ func (ra reqAuthorizor) tokenAuth(f http.HandlerFunc, minPermission models.Permi
...
+// authorize() requires tokenAuth to be called first or will fail
+func (ra reqAuthorizor) authorize(f http.HandlerFunc) http.HandlerFunc {
+	log := func(msg string, args ...interface{}) {
+		log.Printf("authorize: "+msg, args...)
+	}
+	return func(w http.ResponseWriter, r *http.Request) {
+		apikey := models.APIKey{}
+		if ak, ok := r.Context().Value(apiKeyCtxKey).(models.APIKey); ok {
+			apikey = ak
+		}
```
bkeroackdsc, line 178: Same here. Return 403 if APIKey isn't found.

bkeroackdsc (changes requested):
A few small changes I missed in last CR, but otherwise it's looking good.

pkg/api/middleware.go
esterlinej: Adding this to the first thing checked within the middleware handler function

pkg/api/middleware.go
esterlinej: Adding this to the first thing checked within the middleware handler function

pkg/api/middleware.go
esterlinej: Adding this to the first thing checked within the middleware handler function, but cannot include user, because when we forward the root key, no user exists.

bkeroackdsc (approved):

PR #154, opened 2021-03-19
User API Token UI

esterlinej:
### Implement User API Token UI 
- Add stub for User Settings page
 - Add link to User Settings page in user dropdown menu in header
 - Add API UI handler & path support
 - Add API v2 functions & tests for:
   - `GET /v2/user/tokens`
   - `POST /v2/user/token`
   - `DELETE /v2/user/token/{id}`
 ### Changes & Addons
 - Remove `name` field from `api_keys`, update datalayer, api, tests, and ui accordingly
 - Fix API & Fake issues with `DeleteAPIKeyByID`
 - Add API Key limit (_set to 10_) for `apiKeysHandler` and test


esterlinej:
Looking for suggestions on the API Documents help icon (anchored to the API Keys upper right bar), possibly make it into its own group with verbiage, as it doesn't necessarily belong in the API Keys heading group, but did this to keep things compact.

pkg/persistence/apikeys.go
```diff
@@ -25,9 +25,8 @@ func (pg *PGLayer) CreateAPIKey(ctx context.Context, permissionLevel models.Perm
 	in := &models.APIKey{
 		ID:              id,
 		Created:         time.Now().UTC(),
-		LastUsed:        pq.NullTime{Time: time.Now().UTC(), Valid: true},
+		LastUsed:        pq.NullTime{Time: time.Time{}, Valid: false},
```
bkeroackdsc, line 28: If you omit this field it will do the same thing, since the zero value for `pq.NullTime` would have those values (zero value time.TIme, false).

ui/assets/user.js
```diff
@@ -0,0 +1,376 @@
+let apiKeyID = "";
+
+// showHideAPIKeysCreateNew toggles views based on render api keys response
+function showHideAPIKeysCreateNew() {
+    console.log("DEBUG:: showHideAPIKeysCreateNew");
```
bkeroackdsc, line 5: debug console.log

ui/assets/user.js
```diff
@@ -0,0 +1,376 @@
...
+        document.getElementById("apiKeys").setAttribute("class", "row hidden");
+    } else {
+        document.getElementById("noAPIKeysCreateNew").setAttribute("class", "row hidden");
+        document.getElementById("apiKeys").setAttribute("class", "row");
+    }
+}
+
+// setAPIKeys replaces the existing api keys body
+function setAPIKeys(apiKeysBody) {
+    console.log("DEBUG:: setAPIKeys");
```
bkeroackdsc, line 17: debug console.log

ui/assets/user.js
```diff
@@ -0,0 +1,376 @@
...
+        return;
+    }
+    let oldAPIKeysBody = document.getElementById("apiKeysBody");
+    oldAPIKeysBody.parentNode.replaceChild(apiKeysBody, oldAPIKeysBody);
+    showHideAPIKeysCreateNew();
+}
+
+// renderCardHeader populates the api key headers
+function renderCardHeader(apiKey) {
+    console.log("DEBUG:: renderCardHeader");
```
bkeroackdsc, line 28: debug console.log

ui/assets/user.js
```diff
@@ -0,0 +1,376 @@
...
+    apiKeyRow.className = 'row';
+    apiKeyRow.appendChild(apiKeyCol);
+    apiKeyRow.appendChild(deleteAPIKeyCol);
+
+    return apiKeyRow
+}
+
+// renderCardBodyTableBodyRow populates the table within the card body
+function renderCardBodyTableBodyRow(heading, value){
+    console.log("DEBUG:: renderCardBodyTableBodyRow");
```
bkeroackdsc, line 97: debug console.log

ui/assets/user.js
```diff
@@ -0,0 +1,376 @@
...
+    let tr = document.createElement("tr");
+    tr.appendChild(th);
+    tr.appendChild(td);
+
+    return tr
+}
+
+// permissionLevel returns permission level mapping of integer:string
+function permissionLevel() {
+    console.log("DEBUG:: permissionLevel");
```
bkeroackdsc, line 115: debug console.log

bkeroackdsc (changes requested):

pkg/api/api_v2.go
```diff
@@ -1082,3 +1087,152 @@ func (api *v2api) userEnvPodLogsHandler(w http.ResponseWriter, r *http.Request)
...
+		return
+	}
+	reqParams := v2CreateUserAPIKey{}
+	err = json.NewDecoder(r.Body).Decode(&reqParams)
+	if err != nil {
+		api.rlogger(r).Logf("error unmarshaling body: %v", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	if reqParams.Permission < 0 || reqParams.Permission < models.MinPermissionLevel || reqParams.Permission > models.MaxPermissionLevel {
```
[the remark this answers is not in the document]
esterlinej: Yes, thank you for reminding me, I meant to ask if we should be blocking Admin key creation at the API level, or enforce admin key `/user/token/admin` to require a root key to process that request. But that being the case why not just use the root key.

pkg/api/api_v2.go
```diff
@@ -1082,3 +1087,152 @@ func (api *v2api) userEnvPodLogsHandler(w http.ResponseWriter, r *http.Request)
 		w.WriteHeader(http.StatusInternalServerError)
 	}
 }
+
+type v2CreateUserAPIKey struct {
+	Permission  int    `json:"permission"`
```
bkeroackdsc, line 1092: This should be the constant type, not a bare integer.

https://github.com/dollarshaveclub/acyl/blob/1c0c673d131ffb3cabc456154c2508df9b31d135/pkg/models/apikeys.go#L49

pkg/api/api_v2.go
```diff
@@ -1082,3 +1087,152 @@ func (api *v2api) userEnvPodLogsHandler(w http.ResponseWriter, r *http.Request)
...
+	id := mux.Vars(r)["id"]
+	if id == "" {
+		api.rlogger(r).Logf("error token id required")
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+	keyID, err := uuid.Parse(id)
+	if err != nil {
+		api.rlogger(r).Logf("error parsing user token: %v", err)
+		w.WriteHeader(http.StatusInternalServerError)
```
bkeroackdsc, line 1214: This should be a 403 Bad Request response code.

pkg/api/api_v2.go
```diff
@@ -1082,3 +1087,152 @@ func (api *v2api) userEnvPodLogsHandler(w http.ResponseWriter, r *http.Request)
...
+		return
+	}
+	if uis.GitHubUser != apikey.GitHubUser {
+		api.rlogger(r).Logf("action not authorized by user: %v", uis.GitHubUser)
+		w.WriteHeader(http.StatusUnauthorized)
+		return
+	}
+	err = api.dl.DeleteAPIKeyByID(context.Background(), apikey.ID)
+	if err != nil {
+		api.logger.Printf("error deleting user token: %v: %v", id, err)
```
bkeroackdsc, line 1235: You probably should return a 500 error here.

pkg/models/apikeys.go
```diff
@@ -54,3 +53,9 @@ const (
 	WritePermission
 	AdminPermission
 )
+
+const (
+	MinPermissionLevel = int(ReadOnlyPermission)
+	MaxPermissionLevel = int(WritePermission)
+	MaxAPIKeysLimit    = 10
+)
```
bkeroackdsc, line 61: These should be defined in the API package, and should be named like `MinAPIKeyPermissionLevel` etc.

ui/assets/user.js
```diff
@@ -0,0 +1,402 @@
+let apiKeyID = "";
+
+// showHideAPIKeysCreateNew toggles views based on render api keys response
+function showHideAPIKeysCreateNew() {
+    if (!document.getElementById("apiKeysBody").hasChildNodes()) {
```
bkeroackdsc, line 5: For readability, it's preferred to have a positive condition first.

Instead of `if (!foo) { a } else { b }` it's better to do the equivalent: `if (foo) { b } else { a }`.

ui/assets/user.js
```diff
@@ -0,0 +1,402 @@
...
+    }
+}
+
+// encodePermissionLevel encodes the permission level string to integer
+function permissionLevelEncode(permission) {
+    let level = 1;
+    if (permission === "Write") {
+        level = 2;
+    }
+    return level;
```
bkeroackdsc, line 127:
```javascript
if (permission === "Write") {
   return 2;
}
return 1;
```

ui/assets/user.js
```diff
@@ -0,0 +1,402 @@
...
+    let permission = document.getElementById("newAPIKeyPermissionSelect").value;
+    if (permission !== undefined) {
+        if (description === "") {
+            description = "n/a";
+        }
+        apiKeyObj = {
+            "description": description,
+            "permission": permissionLevelEncode(permission),
+        };
+        console.log(`DEBUG: apiKeyObj: ${apiKeyObj.description}`)
```
bkeroackdsc, line 202: Debug printf

ui/assets/user.js
```diff
@@ -0,0 +1,402 @@
...
+});
+
+function getAPIKeys() {
+    let req = new XMLHttpRequest();
+
+    req.open('GET', `${apiBaseURL}/v2/user/tokens`, true);
+    req.onload = function () {
+        let data = [];
+        if (req.status !== 200) {
+            console.log(`user tokens request failed: ${req.status}: ${req.responseText}`);
```
bkeroackdsc, line 329: missing a return

bkeroackdsc (changes requested):

bkeroackdsc (approved):

PR #156, opened 2021-03-22
Refactor API Key Data Model

esterlinej:
### Refactor API Key Data Model
Alter the API Key data model, defining the current key API Key ID as the record ID, and add a column for the Token ID. This will allow us to easily manage keys, deleting from the UI to specify by API Key record ID, without exposing the actual Token ID.

**Add/Update the following:**
- Migrations
- DB
- Models
- API
- Tests
---
### Changes
 - Add token column to api_keys table
 - Update middleware, models, db api key and datalayer functions, fakes, tests


migrations/0022_add_api_keys_token_up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE api_keys ADD COLUMN token UUID;
```
bkeroackdsc, line 1: I would add `NOT NULL DEFAULT gen_random_uuid()`, even though in the code you'll always insert with a value this will ensure that null values never exist.

pkg/persistence/apikeys.go
```diff
@@ -26,19 +30,20 @@ func (pg *PGLayer) CreateAPIKey(ctx context.Context, permissionLevel models.Perm
...
 		return uuid.Nil, errors.Wrap(err, "error inserting api key")
 	}
-	return id, nil
+	return token, nil
 }
 
-// GetAPIKeyById returns the api key object by api key id, nil if not found
-func (pg *PGLayer) GetAPIKeyById(ctx context.Context, id uuid.UUID) (*models.APIKey, error) {
+// GetAPIKey returns the api key by token, nil if not found
+func (pg *PGLayer) GetAPIKey(ctx context.Context, token uuid.UUID) (*models.APIKey, error) {
```
bkeroackdsc, line 43: Rename to `GetAPITokenByToken()`, it stutters but you'll also need a `GetAPITokenByID()` method as well so you want to differentiate between them.

pkg/persistence/apikeys.go
```diff
@@ -69,17 +74,17 @@ func (pg *PGLayer) GetAPIKeysByGithubUser(ctx context.Context, githubUser string
...
+	q := `UPDATE api_keys SET last_used = $1 WHERE token = $2;`
+	_, err := pg.db.ExecContext(ctx, q, pq.NullTime{Time: time.Now().UTC(), Valid: true}, token)
 	if err != nil {
 		return errors.Wrap(err, "error updating api key")
 	}
 	return nil
 }
 
-// DeleteAPIKey unconditionally deletes the api key for the api key id
+// DeleteAPIKey unconditionally deletes the api key for the api key record id
```
bkeroackdsc, line 87: Rename to `DeleteAPIKeyByID()`.

bkeroackdsc (changes requested):

pkg/persistence/apikeys.go
esterlinej: I was thinking about this last night, will update

bkeroackdsc (approved):

wwalexander:
LGTM

PR #160, opened 2021-06-09
Upgrade to Helm 3

### LINKED ISSUE #44: Helm v3 support
Now that Helm v3 is in public beta, we should start to consider how we will support v3 and what changes are needed. At the very least, not needing Tiller should speed up initial environment creation.

From a security perspective, we could come up with a reduced permissions way of installing/upgrading charts since we only need the ClusterAdmin role for creating the initial namespace. 

https://helm.sh/blog/helm-v3-beta/

Will also solve #38

esterlinej:
### Features & Upgrades
- update modules
- utilize metahelm v7.0+ Helm 3 update
- update acyl packages
- remove support for Tiller
- update commands with tiller integration

Fixes #44


bkeroackdsc:
@esterlinej  You're on the right track but I don't think you can share a `metahelm.Manager` as a field on ChartInstaller.

`metahelm.Manager` is threadsafe, but it contains an embedded `*action.Configuration` which should be customized to a specific environment as it's being built. Specifically, each time a new namespace is created, acyl needs to create a new `*action.Configuration` with the namespace and driver info, and then must run `cfg.Init(...)` on that configuration before creating a new `metahelm.Manager` and then using that to install/upgrade all the charts for the environment.

The way I would approach this is instead of having a literal `*metahelm.Manager` field on the ChartInstaller struct, instead have something like:
```go
type ChartInstaller struct {
// ...
  mhmf     func(...) *metahelm.Manager
// ...
}
```
In the constructor when you create a ChartInstaller, you can inject in a function that either creates a real manager or a mock one or whatever you need.

In recent Helm v3 versions, `cfg.Init()` (on `*action.Configuration`) is [threadsafe](https://github.com/helm/helm/issues/9089), so when a new environment is being built, after the namespace has been created and set up, ChartInstaller will need to run something like `cfg.Init(kc, "nitro-9999-my-namespace", <driver name>, nil)`.

esterlinej:
Even though we're defining the kubernetes namespace (`metahelm.WithK8sNamespace(namespace)`), and Acyl creating the namespace per usual, metahelm appears not to be respecting that namespace, or at least not setting it within the generated the context for the `*action.Configuration`.

I found a way to enforce this within Acyl, but wondering if this shouldn't be automatically handled at the metahelm package itself, than forcing it from Acyl, since we already provide the namespace to metahelm in the initialization. I'll continue to investigate, and hopefully find a better solution, though what I have currently is now installing the charts into the expected namespace.

esterlinej:
TODO:
- Official Metahelm Helm3 (3.6.0) build, updating Acyl module dependency
- Thorough Functional Testing QAEs (Create, Update, Destroy)
- Deployment Strategy, possibly move DB changes & migration to it's own branch

pkg/nitro/metahelm/metahelm.go
```diff
@@ -382,24 +406,23 @@ func (ci ChartInstaller) BuildAndInstallCharts(ctx context.Context, newenv *EnvI
...
 }
 
-func (ci ChartInstaller) installOrUpgradeCharts(ctx context.Context, taddr, namespace string, csl []metahelm.Chart, env *EnvInfo, b images.Batch, upgrade bool) error {
+func (ci ChartInstaller) installOrUpgradeCharts(ctx context.Context, namespace string, csl []metahelm.Chart, env *EnvInfo, b images.Batch, upgrade bool) error {
 	eventlogger.GetLogger(ctx).SetK8sNamespace(namespace)
+	mhm, err := ci.mhmf(ctx, ci.kc, ci.hccfg, namespace)
+	if err != nil {
+		return fmt.Errorf("error getting helm client configuration: %w", err)
+	}
+	if mhm == nil {
```
bkeroackdsc, line 423: Add this to the previous conditional: `if err != nil || mhm == nil { ... }`

esterlinej:
Running functional tests, but this is basically code complete, and ready for official review.

bkeroackdsc (changes requested):

bkeroackdsc (approved):

PR #163, opened 2021-09-08
Overhaul integration tests

bkeroackdsc:
- Run integration tests in minikube on demand instead of using an external EKS cluster
- Remove static github token (non-GitHub App) integration tests
- Update palantir/go-githubapp, add longer client timeouts
- Add debug logging for the helm v3 client
- Fix dependabot security alert for ulikunitz/xz


esterlinej (approved):
Looks good, I like the simplification on metahelm package.

